PR #51, opened 2017-03-16
Mlibz 1686 crash on logout

misha-vakulich:
#### Description
App was crashing on Realm acquiring new instance of realm, while last one was no cleared

#### Changes
Add control over realm instances, add finally clauses for most of realm operations

#### Tests
Manual Tests

### DIFF
--- a/android-lib/src/main/java/com/kinvey/android/store/UserStore.java
+++ b/android-lib/src/main/java/com/kinvey/android/store/UserStore.java
@@ -382,7 +382,7 @@ public class UserStore {
      * @param client {@link Client} an instance of the client
      */
     public static void logout(AbstractClient client, KinveyClientCallback<Void> callback) {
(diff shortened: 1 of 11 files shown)


android-lib/src/main/java/com/kinvey/android/store/UserStore.java
```diff
@@ -422,6 +422,13 @@ public void keepOfflineStorageOnLogout(){
     }
 
     /**
+     * Get a flag which allow local offline storage to persist after calls to logout.
+     */
+    public boolean isKeepOfflineStorageOnLogout() {
```
vinaygahlawat: @misha-vakulich How does this `clearStorage` flag set?  I see that there is a default value of `true`, and a method to set it explicitly to `false`, but I don't see a way of toggling this value.

PR #54, opened 2017-04-19
migrate constructors from 2.x ver

mvakulich:
#### Description
new constructor to pass properties
#### Changes
new constructor to pass properties
#### Tests
manual

### DIFF
--- a/android-lib/src/main/java/com/kinvey/android/Client.java
+++ b/android-lib/src/main/java/com/kinvey/android/Client.java
@@ -478,7 +478,7 @@ public class Client extends AbstractClient {
          * @param context - Your Android Application Context
          *
          */
-        public Builder(HttpTransport transport, Context context) {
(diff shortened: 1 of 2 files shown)


android-lib/src/main/java/com/kinvey/android/Client.java
```diff
@@ -475,72 +476,109 @@ public Builder(String appKey, String appSecret, Context context, HttpTransport t
          * </p>
          *
          * @param context - Your Android Application Context
-         * @param transport - Your HttpTransport
          *
          */
-        public Builder(Context context, HttpTransport transport) {
+        public Builder(HttpTransport transport, Context context) {
```
vinaygahlawat: @misha-vakulich Is this change to this `Builder` constructor a breaking change from the last beta release?  If so, please make a note of it to mention this in the change log.

PR #57, opened 2017-05-11
Mlibz 1681 set custom user fields on sighup

misha-vakulich:
#### Description
allow user to provide custom fields for user class
#### Changes
allow user to provide custom fields for user class

#### Tests
jUnit


java-lib/src/com/kinvey/nativejava/Client.java
```diff
@@ -23,8 +23,8 @@
 import com.google.api.client.json.JsonFactory;
 import com.google.api.client.json.JsonObjectParser;
 import com.google.common.base.Preconditions;
-import com.kinvey.java.AbstractClient;
-import com.kinvey.java.Logger;
+import com.kinvey.java.*;
```
vinaygahlawat: Why do we need a wildcard import statement here?

PR #74, opened 2017-06-08
fix: MLIBZ-1893 File can't be downloaded if its size is zero

yuliya-guseva:
#### Description
File couldn't be downloaded if its size is zero. Got '416 Requested range not satisfiable' and the download was looping because of retrying to send new requests.

#### Changes
Check metadata size and don't add range header if it's empty.

#### Tests
JUnit

### DIFF
--- a/java-api-core/src/com/kinvey/java/core/MediaHttpDownloader.java
+++ b/java-api-core/src/com/kinvey/java/core/MediaHttpDownloader.java
@@ -270,7 +270,7 @@ public class MediaHttpDownloader {
             }
 
             // set Range header (if necessary)
-            if (bytesDownloaded != 0 || currentRequestLastBytePos != -1) {
+            if (metaData.getSize() != 0 && (bytesDownloaded != 0 || currentRequestLastBytePos != -1)) {
                 StringBuilder rangeHeader = new StringBuilder();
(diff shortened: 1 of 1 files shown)


java-api-core/src/com/kinvey/java/core/MediaHttpDownloader.java
```diff
@@ -270,7 +270,7 @@ public FileMetaData download(FileMetaData metaData, OutputStream out) throws IOE
             }
 
             // set Range header (if necessary)
-            if (bytesDownloaded != 0 || currentRequestLastBytePos != -1) {
+            if (metaData.getSize() != 0 && (bytesDownloaded != 0 || currentRequestLastBytePos != -1)) {
```
vinaygahlawat: @yuliya-guseva If the file size is zero, wouldn't the `bytesDownloaded` also be zero?

vinaygahlawat: Per conversation with Yuliya:
"if the file is empty, we’re getting bytesDownloaded = 0, but currentRequestLastBytePos != -1 because chunkSize != 0, that’s why it gets into the if. But we can’t simplify it to: if (bytesDownloaded != 0 && currentRequestLastBytePos != -1) , for example, because it won’t be working for non-empty files"

PR #80, opened 2017-06-22
feat: MLIBZ-1926 Research File Descriptor issues in Android tests

yuliya-guseva:
#### Description
Fix memory leaks

#### Changes
Terminate HandlerThread after tests, close stream, remove files after tests

#### Tests
JUnit

### DIFF
--- a/android-lib/src/main/java/com/kinvey/android/Client.java
+++ b/android-lib/src/main/java/com/kinvey/android/Client.java
@@ -997,9 +997,11 @@ public class Client extends AbstractClient {
     }
 
     public void stopKinveyHandlerThread() {
-        kinveyHandlerThread.stopHandlerThread();
-        kinveyHandlerThread.quit();
-        kinveyHandlerThread.interrupt();
+        if (kinveyHandlerThread != null) {
+            kinveyHandlerThread.stopHandlerThread();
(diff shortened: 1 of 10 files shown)


android-lib/src/main/java/com/kinvey/android/KinveyHandlerThread.java
```diff
@@ -38,4 +39,16 @@ protected synchronized void onLooperPrepared() {
         }
     }
 
+    public synchronized void stopHandlerThread() {
```
vinaygahlawat: @yuliya-guseva Please see if we can reduce the access level of this method to package-level.

android-lib/src/main/java/com/kinvey/android/Client.java
```diff
@@ -995,5 +995,13 @@ public FileStore getFileStore(StoreType storeType) {
     protected ICacheManager getSyncCacheManager() {
         return syncCacheManager;
     }
+
```
vinaygahlawat: @yuliya-guseva Please document this method, in terms of why a developer would call it.

yuliya-guseva: Done

PR #82, opened 2017-06-30
feat: MLIBZ-1834 Add client ID to loginWithMic call (Android)

yuliya-guseva:
#### Description
Add client ID to loginWithMic call
https://kinvey.atlassian.net/browse/MLIBZ-1834

#### Changes
Added client ID to login with MIC, added Unit tests

#### Tests
Unit tests


yuliya-guseva:
@vinaygahlawat 
`clientId` needs to be stored in `Credential` to be used in `useRefreshToken()` (to get new access token if old token was expired). `useRefreshToken()` is used in `AbstractKinveyClientRequest` in case the response has code 401. Customers don't have access to `AbstractKinveyClientRequest` and can't set manually `clientId` there. Getting `clientId` from credentials is slower than getting it from `Client` and I've seen this logic in iOS Client: https://github.com/Kinvey/swift-sdk/blob/master/Kinvey/Kinvey/Client.swift
    ```
internal var clientId: String? {
        willSet {
            keychain.clientId = clientId
        }
    }
```

vinaygahlawat:
I agree that `clientId` has to be stored in `Credential`, because it needs to be persisted as well as used in the refresh token logic that is in `useRefreshToken()`.

But in terms of getting `clientId` from `Client` instead of `Credential`, I think we should favor getting it from `Credential`, because it eliminates unnecessary state from being managed.  Even though access to `clientId` may be slower via `Credential`, it would be better to prefer maintainability over speed, since this operation happens infrequently.

yuliya-guseva:
@vinaygahlawat I did more investigations on keeping clientID only on Credential. 
Currently, Client.clientId is used for keeping clientID while a user navigates between LoginActivity and android-kinvey-sdk, also for keeping clientID for using in UserStore and UserStoreRequestManager, if I remove Client.clientID field I will need to add clientId as a parameter in some methods in UserStore (getMICAccessToken, onOAuthCallbackRecieved) and UserStoreRequestManager and I need some time for it and for checking this improvement. 
It is a bit complicated logic for MICLogin in UserStore and UserStoreRequestmanager.

PR #83, opened 2017-07-07
fix: MLIBZ-1941 Cannot resolve 'UserStore.update()' method

yuliya-guseva:
#### Description
Make update method not static

#### Changes
Update method is now in User class and is not static, added tests

#### Tests
Instrumented


android-lib/src/main/java/com/kinvey/android/Client.java
```diff
@@ -129,14 +129,28 @@ protected Client(HttpTransport transport, HttpRequestInitializer httpRequestInit
...
         syncCacheManager = new RealmCacheManager("sync_", this);
     }
 
     public static Client sharedInstance(){
-    	return _sharedInstance;
+    	return sharedInstance;
+    }
+
+    @Override
+    public void setUser(T user) {
```
vinaygahlawat: @yuliya-guseva This method should probably be called `setActiveUser()`, to pair it with `getActiveUser()`.

PR #85, opened 2017-07-27
fix: MLIBZ-1945 Android SDK sends empty query object on get all

yuliya-guseva:
#### Description
Don't send empty query object

#### Changes
Remove sending empty query object

#### Tests
Instrumented

### DIFF
--- a/java-api-core/src/com/kinvey/java/network/NetworkManager.java
+++ b/java-api-core/src/com/kinvey/java/network/NetworkManager.java
@@ -506,8 +506,7 @@ public class NetworkManager<T extends GenericJson> {
 
         MetadataGet(DeltaGet getRequest){
             super(client, "GET", REST_PATH, null, DeltaSetItem[].class);
-            String queryFilterString = getRequest.queryFilter;
-            this.queryFilter = !queryFilterString.equals("{}") ? queryFilterString : null;
+            this.queryFilter = getRequest.queryFilter;
 
             this.skip = getRequest.skip;
             this.limit = getRequest.limit;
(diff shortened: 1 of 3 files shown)


java-api-core/src/com/kinvey/java/network/NetworkManager.java
```diff
@@ -679,7 +678,7 @@ public Aggregate aggregate(ArrayList<String> fields, AggregateEntity.AggregateTy
         Get(String queryString, Class myClass){
         	super(client, "GET", REST_PATH, null, myClass);
         	this.collectionName= NetworkManager.this.collectionName;
-        	this.queryFilter = queryString;
+        	this.queryFilter = !queryString.equals("{}") ? queryString : null;
```
vinaygahlawat: @yuliya-guseva Is it possible here that `queryString` can be null?  If so, we should guard against that case.

java-api-core/src/com/kinvey/java/query/AbstractQuery.java
```diff
@@ -102,6 +102,10 @@ public String getQueryFilterJson(JsonFactory factory) {
             jsonResult = writer.toString();
         } catch (Exception ex) {}
 
+        if (jsonResult.equals("{}")) {
```
vinaygahlawat: Since there is a case above where we are checking against `{}`, do we need to set `jsonResult` to `null` here?

java-api-core/src/com/kinvey/java/network/NetworkManager.java
yuliya-guseva: It's impossible, because before calling "Get(String queryString, Class myClass)" the queryString is checked for null in precondition.

java-api-core/src/com/kinvey/java/query/AbstractQuery.java
yuliya-guseva: It's need because this case doesn't depend on the case above. When we set it to null then the empty query won't be added.

PR #86, opened 2017-07-27
fix: MLIBZ-1947 On logout, the Realm instance is still running and the logout attempt fails.

yuliya-guseva:
#### Description
Logout fails because Realm wasn't closed

#### Changes
Close Realm if there're some exceptions

#### Tests
Instrumented

### DIFF
--- a/android-lib/src/main/java/com/kinvey/android/cache/RealmCache.java
+++ b/android-lib/src/main/java/com/kinvey/android/cache/RealmCache.java
@@ -17,7 +17,6 @@
 package com.kinvey.android.cache;
 
 import com.google.api.client.json.GenericJson;
-import com.kinvey.java.KinveyException;
 import com.kinvey.java.Query;
 import com.kinvey.java.cache.ICache;
 import com.kinvey.java.query.AbstractQuery;
--- a/android-lib/src/main/java/com/kinvey/android/cache/RealmCache.java
+++ b/android-lib/src/main/java/com/kinvey/android/cache/RealmCache.java
@@ -17,6 +17,7 @@
 package com.kinvey.android.cache;
 
 import com.google.api.client.json.GenericJson;
+import com.kinvey.java.KinveyException;
 import com.kinvey.java.Query;
 import com.kinvey.java.cache.ICache;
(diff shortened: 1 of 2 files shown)


yuliya-guseva:
@vinaygahlawat 
I've investigated it and found out in the documentation that it's important to cancel the transaction if an exception was caught: https://realm.io/docs/java/latest/#writes

 "Realm is crash safe, so in case of an Exception happening inside a transaction, the Realm itself will not be corrupted. Only the data in the current transaction will be lost. In case the Exception is caught and the app continues, it is important to cancel the transaction."

 As well I found out the question on the GitHub: https://github.com/realm/realm-java/issues/4217
 There is example of safe using beginTransaction and commitTransaction in the comments:
 "try {
    realm.beginTransaction();
    // body of executeTransaction
    realm.commitTransaction();
} catch(Exception e) { 
    if(realm.isInTransaction()) {
         realm.cancelTransaction();
    }
    throw new RuntimeException(e);
}"

But we use logic like this:
"try{
    realm.beginTransaction();
	....
    realm.commitTransaction();
} finally {
    realm.close();
}"

And realm has in the documentation an example where they don't catch exception when they close realm in finally block
https://realm.io/docs/java/latest/#realm-threading-example

```
public class PollingService extends IntentService {
    @Override
    public void onHandleIntent(Intent intent) {
        Realm realm = Realm.getDefaultInstance();
        try {
            // go do some network calls/etc and get some data and stuff it into a 'json' var
            String json = customerApi.getCustomers();
            realm.beginTransaction();
            realm.createObjectFromJson(Customer.class, json); // Save a bunch of new Customer objects
            realm.commitTransaction();
            // At this point, the data in the UI thread is already up to date.
            // ...
        } finally {
            realm.close();
        }
    }
    // ...
}
```

Therefore I think when we are using `mRealm.close();` we don't need to do `realm.cancelTransaction();` .

vinaygahlawat:
Sounds good, thanks.  And maybe in the future, we can investigate the merits of using `executeTransaction()` instead of manually managing transactions with `beginTransaction()` and `commitTransaction()`.

https://realm.io/docs/java/latest/#transaction-blocks

PR #88, opened 2017-08-04
feat: MLIBZ-1993 Sync Count Not Impelemented for DataStore

yuliya-guseva:
#### Description
Add syncCount method in DataStore

#### Changes
Added syncCount method, added Instrumented test for this method

#### Tests
Instrumented

### DIFF
--- a/android-lib/src/main/java/com/kinvey/android/store/DataStore.java
+++ b/android-lib/src/main/java/com/kinvey/android/store/DataStore.java
@@ -757,6 +757,14 @@ public class DataStore<T extends GenericJson> extends BaseDataStore<T> {
         return new Query(new MongoQueryFilter.MongoQueryFilterBuilder());
     }
 
+    /**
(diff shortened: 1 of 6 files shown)


android-lib/src/main/java/com/kinvey/android/store/DataStore.java
```diff
@@ -757,6 +757,14 @@ public Query query() {
         return new Query(new MongoQueryFilter.MongoQueryFilterBuilder());
     }
 
+    /**
+     * This methods gets a count of entities modified locally and pending a push to the backend
+     * @return the count of sync objects for given collection
+     */
+    public long syncCount() {
+        return client.getSycManager().getCount(getCollectionName());
```
vinaygahlawat: @yuliya-guseva While we are adding this, let's also correct the spelling of the `getSycManager()` method.

PR #91, opened 2017-08-16
feat: MLIBZ-1868 Travis integration for Android repo

yuliya-guseva:
#### Description
Configure Travis, add Code Coverage

#### Changes
Configure Travis, add Code Coverage

#### Tests
Instrumented


android-lib/src/androidTest/assets/kinvey.properties
```diff
@@ -1,6 +1,6 @@
 # Required settings
-app.key=<YOUR_APP_KEY_HERE>
-app.secret=<YOUR_APP_SECRET_HERE>
+app.key=kid_BkssrV2eW
```
tejasranade: @yuliya-guseva do these keys need to be included here, or is it an accident? We prefer not to have any keys in public, so I suggest removing these from here and adding them as environment variables if necessary.

yuliya-guseva: These keys are needed, they're used in tests. Sure, moving it to Travis environment variables.

yuliya-guseva:
@tejasranade Please check the latest changes, now it doesn't fail.

PR #92, opened 2017-08-21
feat: MLIBZ-2033 Delta Set Caching not working correctly

yuliya-guseva:
#### Description
Fix saving kmd to cache

#### Changes
Fix saving kmd to cache and add tests

#### Tests
Instrumented

### DIFF
--- a/android-lib/src/main/java/com/kinvey/android/cache/ClassHash.java
+++ b/android-lib/src/main/java/com/kinvey/android/cache/ClassHash.java
@@ -21,7 +21,6 @@ import com.google.api.client.util.ClassInfo;
 import com.google.api.client.util.Data;
 import com.google.api.client.util.FieldInfo;
 import com.kinvey.java.model.KinveyMetaData;
-import com.kinvey.java.store.BaseFileStore;
 
 import java.lang.reflect.Array;
 import java.lang.reflect.Field;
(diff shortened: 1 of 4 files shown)


android-lib/src/main/java/com/kinvey/android/cache/ClassHash.java
```diff
@@ -217,12 +217,24 @@ public static DynamicRealmObject saveData(String name, DynamicRealm realm, Class
         if (obj.containsKey("_kmd")){
             Map kmd = (Map)obj.get("_kmd");
             if (kmd != null) {
-                kmd.put("lmt", String.format("%tFT%<tTZ",
-                        Calendar.getInstance(TimeZone.getTimeZone("Z"))));
+                KinveyMetaData metadata = new KinveyMetaData();
```
tejasranade: A cleaner way to do this is to expose a static method on the `KinveyMetaData` class that constructs the class from a `Map`. So you can call something like - 
`KinveyMetaData.buildFromMap(kmd)`.

This encapsulates the keys and any other logic for dates etc. within the `KinveyMetadata` class.

yuliya-guseva: Done

PR #93, opened 2017-08-28
feat: MLIBZ-2006 Implement encryption API on Android

yuliya-guseva:
#### Description
Implement encryption API

#### Changes
Added Client#Builder#setEncryptionKey method, added Instrumented test for this method

#### Tests
Instrumented


android-lib/src/main/java/com/kinvey/android/cache/RealmCacheManager.java
```diff
@@ -190,17 +204,29 @@ private String getClientHash(){
     DynamicRealm getDynamicRealm(){
         synchronized (LOCK){
             Uri server = Uri.parse(client.getBaseUrl());
-            DynamicRealm realm = DynamicRealm.getInstance(getRealmConfiguration());
+            DynamicRealm realm;
+            try {
+                realm = DynamicRealm.getInstance(getRealmConfiguration());
+            } catch (RealmFileException exception) {
+                throw new KinveyException("Access Error", "Use correct encryption key", "You are using wrong encryption key");
```
vinaygahlawat: Is it possible to catch another `RealmFileException` that is not an error with encryption key? If so, do we need to inspect the exception before reporting as an encryption key exception?

android-lib/src/main/java/com/kinvey/android/cache/RealmCacheManager.java
```diff
@@ -54,6 +56,18 @@
...
+        this.client = client;
+        this.context = client.getContext();
+    }
+
+    public RealmCacheManager(byte[] encryptionKey, String prefix, Client client){
+        this.encryptionKey = encryptionKey;
+        this.client = client;
+        this.context = client.getContext();
+    }
+
```
vinaygahlawat: What is the use of `prefix` in this constructor?

android-lib/src/main/java/com/kinvey/android/Client.java
```diff
@@ -125,14 +126,15 @@
     protected Client(HttpTransport transport, HttpRequestInitializer httpRequestInitializer, String rootUrl,
                      String servicePath, JsonObjectParser objectParser,
                      KinveyClientRequestInitializer kinveyRequestInitializer, CredentialStore store,
-                     BackOffPolicy requestPolicy, Context context) {
+                     BackOffPolicy requestPolicy, byte[] encryptionKey, Context context) {
```
vinaygahlawat: Is this a breaking change for this constructor?

android-lib/src/main/java/com/kinvey/android/cache/RealmCacheManager.java
yuliya-guseva: Prefix is used to create realmConfiguration: "realmConfiguration = new RealmConfiguration.Builder().name(prefix + "_" + getClientHash())". At the current moment the prefix isn't saved (wasn't saved previously). It should be investigated more.

android-lib/src/main/java/com/kinvey/android/Client.java
yuliya-guseva: Yes, because we need the encryptionKey in 'Client' to initialize RealmCacheManager during Client initialization process and initialize new RealmCacheManager after logout (because android-sdk is cleared after logout).

android-lib/src/main/java/com/kinvey/android/cache/RealmCacheManager.java
yuliya-guseva: I've seen only this exception, but we can check the RealmFileException's message. If RealmFileException's message has "Access Error" then we will throw KinveyException("Access Error"), in other case the RealmFileException will be thraown.

PR #95, opened 2017-09-12
feat: Add DataStore#save(ListOfEntities) method

yuliya-guseva:
#### Description
Add DataStore#save(ListOfEntities) method

#### Changes
Added new method DataStore.save for saving list of entities

#### Tests
Instrumented

### DIFF
--- a/java-api-core/src/com/kinvey/java/store/requests/data/save/SaveListRequest.java
+++ b/java-api-core/src/com/kinvey/java/store/requests/data/save/SaveListRequest.java
@@ -52,10 +52,18 @@ public class SaveListRequest<T extends GenericJson> implements IRequest<List<T>>
     @Override
     public List<T> execute() throws IOException {
         List<T> ret = new ArrayList<T>();
-        for (T obj : objects){
-            SaveRequest<T> save = new SaveRequest<T>(
-                    cache, networkManager , writePolicy, obj, syncManager);
-            ret.add(save.execute());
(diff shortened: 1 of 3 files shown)


java-api-core/src/com/kinvey/java/store/requests/data/save/SaveListRequest.java
```diff
@@ -52,10 +52,18 @@ public SaveListRequest(ICache<T> cache, NetworkManager<T> networkManager, WriteP
...
+            for (T obj : ret){
+                syncManager.enqueueRequest(networkManager.getCollectionName(),
+                        networkManager.saveBlocking(obj));
+            }
+        } else {
+            for (T obj : objects){
+                SaveRequest<T> save = new SaveRequest<T>(
+                        cache, networkManager , writePolicy, obj, syncManager);
+                ret.add(save.execute());
+            }
```
vinaygahlawat: What happens in the `LOCAL_THEN_NETWORK` case? Does this also make a bulk save of entities to Realm?

yuliya-guseva: In case when write policy is LOCAL_THEN_NETWORK a bulk save won't work. Because there are a bit different logic in LOCAL_THEN_NETWORK case and customer asked only about StoreType.SYNC. I think we can think about improvement this logic for LOCAL_THEN_NETWORK, but I think it should be a different ticket.

PR #101, opened 2017-10-03
feat: MLIBZ-2073 Change logic for Sync Requests saving

yuliya-guseva:
#### Description
Change logic for Sync Requests saving

#### Changes
Removed saving full object to Sync cache 
Moved Sync requests initialization before pushing items to server

#### Tests
Instrumented

### DIFF
--- a/java-api-core/src/com/kinvey/java/sync/RequestMethod.java
+++ b/java-api-core/src/com/kinvey/java/sync/RequestMethod.java
@@ -1,0 +1,25 @@
+package com.kinvey.java.sync;
+
+/**
+ * Created by yuliya on 09/28/17.
+ */
+
+public enum RequestMethod {
+    SAVE("SAVE"),
+    DELETE("DELETE");
+
+    private String query;
+
+    RequestMethod(String query) {
+        this.query = query;
+    }
+
+    public static RequestMethod fromString(String verb){
+        for (RequestMethod v : RequestMethod.values()){
+            if (v.query.equals(verb)){
+                return v;
+            }
+        }
+        return null;
+    }
+}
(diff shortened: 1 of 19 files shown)


java-api-core/src/com/kinvey/java/sync/RequestMethod.java
```diff
@@ -0,0 +1,25 @@
...
+
+    public static RequestMethod fromString(String verb){
+        for (RequestMethod v : RequestMethod.values()){
+            if (v.query.equals(verb)){
+                return v;
+            }
+        }
+        return null;
+    }
+}
```
vinaygahlawat: Should we capture HTTP verbs instead, to make sure we are doing the appropriate action (POST, PUT, DELETE)?

java-api-core/src/com/kinvey/java/sync/dto/SyncRequest.java
```diff
@@ -73,8 +75,16 @@ public static HttpVerb fromString(String verb){
...
 
+    @Key("requestMethod")
+    private String requestMethod;
 
-    public SyncRequest(){};
+    public SyncRequest(){}
+
+    public SyncRequest(RequestMethod requestMethod, SyncMetaData entityID, String collectionName) {
+        this.requestMethod = requestMethod.name();
+        this.id = entityID;
```
vinaygahlawat: What is the reason for putting entityId, customerVersion and customRequestHeaders in the same object? Is this how it was before?

java-api-core/src/com/kinvey/java/sync/RequestMethod.java
yuliya-guseva: I've saved creating request logic. SDK decides which action (POST, PUT, DELETE) to use before pushing, NetworkManager#saveBlocking and NetworkManager#deleteBlocking is used for it in PushRequest and AsyncPushRequest classes.

java-api-core/src/com/kinvey/java/sync/dto/SyncRequest.java
yuliya-guseva: Yes, it was before, I've saved it.

PR #102, opened 2017-10-06
Feature/mlibz 2107 implement automatic pagination

vinaygahlawat:
#### Description
Customer is running into issues trying to pull more than 10k records into the local cache, which has to do with issues surrounding pagination and deletion of the appropriate page in the local cache. All of the data was being deleted from the database during each page write. This change is to offer an auto-pagination option, so that the SDK handles pagination implicitly.

#### Changes
Providing a flag on the `DataStore` to enable auto-pagination.  If this flag is set to `true`, then pagination through all records of a collection will happen implicitly. This fix includes changes to add skip and limit support for cache deleting.

#### Tests
Instrumented tests added.

### DIFF
--- a/android-lib/src/main/java/com/kinvey/android/cache/RealmCache.java
+++ b/android-lib/src/main/java/com/kinvey/android/cache/RealmCache.java
@@ -298,7 +298,9 @@ public class RealmCache<T extends GenericJson> implements ICache<T> {
                         ids.add((String)id.get("_id"));
                     }
                     mRealm.commitTransaction();
-                    this.delete(ids);
+                    if (ids.size() > 0) {
+                        this.delete(ids);
+                    }
                 } else if (skip > 0) {
                     // only skip modifier has been applied, so take a subset of the Realm result set
                     if (skip < result.size()) {
--- a/android-lib/src/main/java/com/kinvey/android/cache/RealmCache.java
+++ b/android-lib/src/main/java/com/kinvey/android/cache/RealmCache.java
@@ -456,19 +456,26 @@ public class RealmCache<T extends GenericJson> implements ICache<T> {
         return ret;
     }
 
+
     @Override
     public long count(Query q) {
         DynamicRealm mRealm = mCacheManager.getDynamicRealm();
         long ret = 0;
         try {
-            if (!isQueryContainsInOperator(q.getQueryFilterMap())) {
+            if (q != null && !isQueryContainsInOperator(q.getQueryFilterMap())) {
                 mRealm.beginTransaction();
                 RealmQuery<DynamicRealmObject> query = mRealm.where(mCollection);
                 QueryHelper.prepareRealmQuery(query, q.getQueryFilterMap());
                 ret = query.count();
                 mRealm.commitTransaction();
             } else {
-                List<T> list = get(q);
+                List<T> list;
+                if (q != null) {
+                    list = get(q);
+                }
+                else {
+                    list = get();
+                }
                 ret = list.size();
             }
         } finally {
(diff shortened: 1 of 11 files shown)


android-lib/src/main/java/com/kinvey/android/store/DataStore.java
```diff
@@ -442,6 +447,9 @@ public void find(KinveyListCallback<T> callback, KinveyCachedClientCallback<List
         new AsyncRequest<List<T>>(this, methodMap.get(KEY_GET_ALL), callback, getWrappedCacheCallback(cachedCallback)).execute();
     }
 
+    public void findCount(KinveyCountCallback callback) {
```
tejasranade: Why `findCount`? Our other SDKs use `count`.

java-api-core/src/com/kinvey/java/store/BaseDataStore.java
```diff
@@ -302,6 +331,9 @@ public void pushBlocking() throws IOException {
         return networkData;
     }
 
+    public void getCountBlocking() throws IOException {
```
tejasranade: Does this method need to be a public method on the `BaseDataStore`?

java-api-core/src/com/kinvey/java/store/requests/data/read/AbstractReadCountRequest.java
```diff
@@ -0,0 +1,96 @@
...
+    public Integer execute() throws IOException {
+        Integer ret = 0;
+        AbstractKinveyJsonClientRequest<KinveyCountResponse> request = null;
+        try {
+            request = countNetwork();
+        } catch (IOException e) {
+            e.printStackTrace();
+        }
+
+        switch (readPolicy){
```
yuliya-guseva: Will be implemented logic for FORCE_LOCAL and LOCAL_THEN_NETWORK readPolisy?

android-lib/src/main/java/com/kinvey/android/cache/RealmCache.java
```diff
@@ -282,9 +282,46 @@ public int delete(Query query) {
...
+                    else {
+                        ret = 0;
+                    }
+                } else {
+                    // no skip or limit applied to query, so delete all results from Realm
+                    result.deleteAllFromRealm();
+                    mRealm.commitTransaction();
+                }
+
             } else {
```
yuliya-guseva: If here will be false? If query will be contained `in` operator? 
`if (!isQueryContainsInOperator(query.getQueryFilterMap()))`

yuliya-guseva:
Looks like, logic with skip and limit should be implemented and in this methods:
`RealmCache#getFirst(Query q)` 
`RealmCache#count(Query q)`

java-api-core/src/com/kinvey/java/store/requests/data/read/AbstractReadCountRequest.java
vinaygahlawat: Yes, I will include this in a new commit.

vinaygahlawat:
We may possibly have to add `skip` and `limit support for `RealmCache#getFirst(Query q)` and `RealmCache#count(Query q)`. Will have to think about this, to see if it makes sense in the context of these methods to allow these modifiers, as well as to check across SDKs for this support. If we do, I think that this is outside the scope of this PR, so we can create separate tasks to address this.

android-lib/src/main/java/com/kinvey/android/store/DataStore.java
vinaygahlawat: Good point, I'll change to `count`.

java-api-core/src/com/kinvey/java/store/BaseDataStore.java
vinaygahlawat: Actually, I will remove this method, as it is not used anymore.

yuliya-guseva:
LGTM

PR #103, opened 2017-10-10
feat: MLIBZ-2091 Implement calculation methods

yuliya-guseva:
#### Description
 Implement calculation methods

#### Changes
Implemented calculations methods in DataStore and added Instrumented tests to them.

#### Tests
Instrumented tests added


java-api-core/src/com/kinvey/java/store/requests/data/CalculationRequest.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @Override
+    public Aggregation.Result[] execute() throws IOException {
+        Aggregation.Result[] ret = null;
+        switch (readPolicy) {
+            case FORCE_LOCAL:
+                ret = getCached();
+                break;
+            case FORCE_NETWORK:
+            case BOTH:
```
vinaygahlawat: Does `getCached()` get called here if the `readPolicy` is `BOTH`?

yuliya-guseva: CachedCallback for StoreType.CACHE is processed in place where CalculationRequest is created https://github.com/Kinvey/java-library/blob/53d5c6ec09cc89d5b0b7ffd439ccee9c99868308/java-api-core/src/com/kinvey/java/store/BaseDataStore.java#L402

PR #104, opened 2017-10-11
Feature/mlibz 2079 query null fields

vinaygahlawat:
#### Description
Creating a query in which a field is compared to `null` caused the query to produce an empty query against that field, yielding incorrect Mongo query string formation.

#### Changes
This was a result of using the `JsonGenerator.serialze()` method to create the Mongo query string. The [documentation for `JsonGenerator`](https://developers.google.com/api-client-library/java/google-http-java-client/reference/1.20.0/com/google/api/client/json/JsonGenerator) points out that if the value trying to be written is `null`, the object will not be serialized. To fix this, the use of `serialize()` was omitted in favor of using the lower-level methods to write values as JSON, such as `writeNull()`, `writeNumber()`, `writeString()`, etc.

#### Tests
Unit tests added to validate generated Mongo query strings.

### DIFF
--- a/android-lib/src/androidTest/java/com/kinvey/androidTest/store/DataStoreTest.java
+++ b/android-lib/src/androidTest/java/com/kinvey/androidTest/store/DataStoreTest.java
@@ -566,13 +566,91 @@ public class DataStoreTest {
 
     @Test
     public void testMongoQueryStringBuilder() {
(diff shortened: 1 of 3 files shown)


yuliya-guseva:
Looks good, but some tests are failed.

PR #106, opened 2017-10-18
MLIBZ-2135 Auto-pagination for blocking pull

vinaygahlawat:
#### Description
The auto-pagination feature was not available for the `pullBlocking` method

#### Changes
Add auto-pagination to `pullBlocking`.

#### Tests
Add integration test.

### DIFF
--- a/android-lib/src/androidTest/java/com/kinvey/androidTest/store/DataStoreTest.java
+++ b/android-lib/src/androidTest/java/com/kinvey/androidTest/store/DataStoreTest.java
@@ -566,13 +566,91 @@ public class DataStoreTest {
 
     @Test
     public void testMongoQueryStringBuilder() {
+        // Arrange
         DataStore<Person> store = DataStore.collection(Person.COLLECTION, Person.class, StoreType.NETWORK, client);
         Query myQuery = client.query();
+        String expectedMongoQuery;
+        String mongoQuery;
+
+        // Act
+        // Assert
+
+        // Test field string value
+        myQuery = client.query();
+        myQuery.equals("testString", "a test");
+        expectedMongoQuery = "{\"testString\":\"a test\"}";
+        mongoQuery = myQuery.getQueryFilterJson(client.getJsonFactory());
+        assertEquals(expectedMongoQuery, mongoQuery);
+
+        // Test field boolean value
+        myQuery = client.query();
+        myQuery.equals("testbool", true);
+        expectedMongoQuery = "{\"testbool\":true}";
+        mongoQuery = myQuery.getQueryFilterJson(client.getJsonFactory());
+        assertEquals(expectedMongoQuery, mongoQuery);
+
+        // Test field int value
+        myQuery = client.query();
+        myQuery.equals("testint", 33);
+        expectedMongoQuery = "{\"testint\":33}";
+        mongoQuery = myQuery.getQueryFilterJson(client.getJsonFactory());
+        assertEquals(expectedMongoQuery, mongoQuery);
+
+        // Test field long value
+        myQuery = client.query();
+        myQuery.equals("testlong", 34L);
+        expectedMongoQuery = "{\"testlong\":34}";
+        mongoQuery = myQuery.getQueryFilterJson(client.getJsonFactory());
+        assertEquals(expectedMongoQuery, mongoQuery);
+
+        // Test field double value
+        myQuery = client.query();
+        myQuery.equals("testdouble", 34.0);
+        expectedMongoQuery = "{\"testdouble\":34.0}";
+        mongoQuery = myQuery.getQueryFilterJson(client.getJsonFactory());
+        assertEquals(expectedMongoQuery, mongoQuery);
+
+        // Test field float value
+        myQuery = client.query();
+        myQuery.equals("testfloat", 34.0f);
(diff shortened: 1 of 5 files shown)


android-lib/src/androidTest/java/com/kinvey/androidTest/store/DataStoreTest.java
```diff
@@ -1187,6 +1342,37 @@ public void testPagedPull() throws InterruptedException {
...
+        save(store, theodore);
+        long cacheSizeBefore = getCacheSize(StoreType.CACHE);
+        assertTrue(cacheSizeBefore == 3);
+        client.getCacheManager().getCache(Person.COLLECTION, Person.class, StoreType.CACHE.ttl).clear();
+        long cacheSizeBetween = getCacheSize(StoreType.CACHE);
+        assertTrue(cacheSizeBetween == 0);
+
+        // Act
+        store.setAutoPagination(true);
+        List<Person> pullResults = store.pullBlocking(null);
```
tejasranade: Even though `autoPagination` is set to `true`, would this paginate on 3 items? Is there a page size that needs to be set?

android-lib/src/main/java/com/kinvey/android/store/DataStore.java
```diff
@@ -682,11 +676,7 @@ public void pull(Query query, KinveyPullCallback<T> callback) {
     public void pull(KinveyPullCallback<T> callback) {
```
tejasranade: Wouldn't it be better for this method to call the other method `pull (query, callback)` with a `null` query? Just to avoid repeated checking of preconditions.

java-api-core/src/com/kinvey/java/query/AbstractQuery.java
```diff
@@ -97,12 +98,16 @@ public String getQueryFilterJson(JsonFactory factory) {
...
         try {
             JsonGenerator generator = factory.createJsonGenerator(writer);
-            generator.serialize(getQueryFilterMap());
+            AbstractMap<String, Object> filterMap = getQueryFilterMap();
+            buildQueryString(generator, filterMap);
             generator.flush();
             jsonResult = writer.toString();
-        } catch (Exception ex) {}
+        } catch (Exception ex) {
+            // TODO add exception handling here instead of suppressing exception
```
tejasranade: What's the resolution of this `TODO`? Should we at least print the exception here?

java-api-core/src/com/kinvey/java/query/AbstractQuery.java
```diff
@@ -111,6 +116,140 @@ public String getQueryFilterJson(JsonFactory factory) {
         return jsonResult;
     }
 
+    private void buildQueryString(JsonGenerator generator, AbstractMap<String, Object> filterMap) throws IOException {
```
tejasranade: How does this relate to `pullBlocking`?

android-lib/src/androidTest/java/com/kinvey/androidTest/store/DataStoreTest.java
vinaygahlawat: In my manual testing of the app, I set the page size to 2 items, to properly test the auto-pagination. In this test, though, this will just verify that this works on a single page. I could make the page size configurable, but I was not sure if it is a good idea to make this available, just for testing purposes.  Should I make the page size configurable?

android-lib/src/main/java/com/kinvey/android/store/DataStore.java
vinaygahlawat: Yeah makes sense.  I was just reverting this back to the way it was, but I'll change it to consolidate the logic.

java-api-core/src/com/kinvey/java/query/AbstractQuery.java
vinaygahlawat: Yes, I will throw an exception here.

java-api-core/src/com/kinvey/java/query/AbstractQuery.java
vinaygahlawat: Interesting.  Actually this does not relate to `pullBlocking`, it is part of another [PR](https://github.com/Kinvey/java-library/pull/104) for querying. I may have accidentally branched this off of the query branch, which is why it is showing up in this PR. If it's ok, I'll merge the other PR, which has already been approved, prior to merging this when its ready. That way, there will be no conflicts.

tejasranade: :+1:

java-api-core/src/com/kinvey/java/query/AbstractQuery.java
tejasranade: :+1:

android-lib/src/androidTest/java/com/kinvey/androidTest/store/DataStoreTest.java
tejasranade: Hmm I see what you mean. It'll be nice to actually test auto-pagination. The page size doesn't need to be publicly available, but see if there's a way we can expose it for the test, or otherwise make the test retrieve multiple pages.

PR #107, opened 2017-10-20
feat: MLIBZ-2132 Implement DataStore#clear

yuliya-guseva:
#### Description
Implemented DataStore#clear method

#### Changes
Added DataStore#clear method for clearing the local cache.

#### Tests
Instrumented, Manual

### DIFF
--- a/java-api-core/src/com/kinvey/java/store/BaseDataStore.java
+++ b/java-api-core/src/com/kinvey/java/store/BaseDataStore.java
@@ -359,7 +359,6 @@ public class BaseDataStore<T extends GenericJson> {
         Preconditions.checkNotNull(client, "client must not be null.");
         Preconditions.checkArgument(client.isInitialize(), "client must be initialized.");
         client.getSyncManager().clear(collection);
-        pullBlocking(null);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


java-api-core/src/com/kinvey/java/store/BaseDataStore.java
```diff
@@ -270,6 +270,15 @@ public T save (T object) throws IOException {
...
     /**
+     * Clear the local cache storage
+     */
+    public void clear() {
+        client.getCacheManager().getCache(getCollectionName(), storeItemType, Long.MAX_VALUE).delete(new Query());
+        client.getSyncManager().getCacheManager().getCache("sync", SyncRequest.class, Long.MAX_VALUE).delete(new Query());
+        client.getSyncManager().getCacheManager().getCache("syncitems", SyncItem.class, Long.MAX_VALUE).delete(new Query());
+    }
+
+    /**
```
vinaygahlawat: I think we should check some preconditions here, just as in the `purge` method. Also, can we just call `purge` from this method, in order to clear out the sync queue?

yuliya-guseva: Makes sense. Preconditions and using purge were added to clear method.

PR #108, opened 2017-10-24
feat: MLIBZ-2140  Collection kmd grows exponentially every time a sync pull is called

yuliya-guseva:
#### Description
Added clearing cache tables (_acl and _kmd) before calling pull/sync

#### Changes
Added clearing cache tables (_acl and _kmd) before calling pull/sync

#### Tests
Instrumented, Manual


yuliya-guseva:
Currently TableManager (the class for getting access to table names) is not public. To add tests we need to make it public. Is it ok it to be changed to public?

vinaygahlawat:
As we discussed, let's not make it public.

PR #109, opened 2017-10-25
fix: MLIBZ-2151 Android sdk 3.0.8 push() is throwing an exception

yuliya-guseva:
#### Description
Fixed BaseDataStore#pushblocking method

#### Changes
Collection name was saved not correctly. It was fixed.

#### Tests
Instrumented, manual.

### DIFF
--- a/java-api-core/src/com/kinvey/java/store/requests/data/PushRequest.java
+++ b/java-api-core/src/com/kinvey/java/store/requests/data/PushRequest.java
@@ -34,14 +34,13 @@ import java.util.List;
  */
 public class PushRequest<T extends GenericJson> extends AbstractKinveyExecuteRequest<T> {
 
-    private final String collectionName;
     private ICache<T> cache;
     private NetworkManager<T> networkManager;
     private final SyncManager syncManager;
     private AbstractClient client;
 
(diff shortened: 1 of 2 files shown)


java-api-core/src/com/kinvey/java/store/requests/data/PushRequest.java
```diff
@@ -34,14 +34,13 @@
...
 
-    private final String collectionName;
     private ICache<T> cache;
     private NetworkManager<T> networkManager;
     private final SyncManager syncManager;
     private AbstractClient client;
 
     public PushRequest(String collectionName, ICache<T> cache, NetworkManager<T> networkManager, AbstractClient client){
-        this.collectionName = collectionName;
+        this.collection = collectionName;
```
AntonDobrev: Not my call, but I am concerned with the usage of fields overriding the `protected` fields of the `abstract` class. At the same time `collection` is setting the derived field from the base class.

- should we think of any problems that may arise with that one
- would it be better to either override fields from the base class or use them in the ctor

Just asking in terms of determining possible implications of that approach in further subclasses of `PushRequest` or `AbstractKinveyExecuteRequest`.

PR #122, opened 2017-12-04
feat: MLIBZ-2143 Add cached callback to DataStore#count

yuliya-guseva:
#### Description
Added cached callback to DataStore#count

#### Changes
Added cached callback to DataStore#count

#### Tests
Instrumented

### DIFF
--- a/java-api-core/src/com/kinvey/java/store/BaseDataStore.java
+++ b/java-api-core/src/com/kinvey/java/store/BaseDataStore.java
@@ -235,12 +235,36 @@ public class BaseDataStore<T extends GenericJson> {
         return find((KinveyCachedClientCallback<List<T>>)null);
     }
 
+    /**
+     * Get items count in collection
+     * @return items count in collection
+     */
     public Integer count() throws IOException {
         Preconditions.checkNotNull(client, "client must not be null.");
         Preconditions.checkArgument(client.isInitialize(), "client must be initialized.");
+        return count(null);
+    }
+
+    /**
+     * Get items count in collection
+     * @param cachedCallback is using with StoreType.CACHE to get items count in collection
+     * @return items count in collection
+     */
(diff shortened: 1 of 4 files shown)


java-api-core/src/com/kinvey/java/store/BaseDataStore.java
```diff
@@ -235,12 +235,36 @@ public T find (String id) throws IOException {
...
+     * @return items count in collection
+     */
+    public Integer count(KinveyCachedClientCallback<Integer> cachedCallback) throws IOException {
+        Preconditions.checkNotNull(client, "client must not be null.");
+        Preconditions.checkArgument(client.isInitialize(), "client must be initialized.");
+        Preconditions.checkArgument(cachedCallback == null || storeType == StoreType.CACHE, "KinveyCachedClientCallback can only be used with StoreType.CACHE");
+        if (storeType == StoreType.CACHE && cachedCallback != null) {
+            Integer ret = new ReadCountRequest<T>(cache, networkManager, ReadPolicy.FORCE_LOCAL, null, client.getSyncManager()).execute();
+            cachedCallback.onSuccess(ret);
+        }
```
vinaygahlawat: The base class for `ReadCountRequest<T>` calls both `countCached()` for the local result and `request.execute().getCount()` for the network result, which means that for `StoreType.CACHE`, we will be calling `countCached()` twice. Should we change the base class implementation for the `ReadPolicy.BOTH` read policy, or should we change the second `ReadCountRequest<T>` to use `ReadPolicy.FORCE_NETWORK`?

yuliya-guseva: I think it will be better to remove `ret = countCached()` from ReadPolicy.BOTH because the result of `countCached()` isn't used.

PR #125, opened 2017-12-07
feat: MLIBZ-2134 Fix sync/pull with skip/limit in a query

yuliya-guseva:
#### Description
Not correct saving items to the cache if sync/pull methods are used with skip/limit parameters in a query. It happens because items are cleared not in correct order before sync/pull.

#### Changes
Added sort by `_kmd.ect` for getting items from the cache.

#### Tests
Instrumented

### DIFF
--- a/android-lib/src/androidTest/java/com/kinvey/androidTest/store/DataStoreTest.java
+++ b/android-lib/src/androidTest/java/com/kinvey/androidTest/store/DataStoreTest.java
@@ -1035,6 +1035,21 @@ public class DataStoreTest {
         return callback;
     }
 
+    private DefaultKinveySyncCallback sync(final DataStore<Person> store, final Query query, int seconds) throws InterruptedException {
+        final CountDownLatch latch = new CountDownLatch(1);
+        final DefaultKinveySyncCallback callback = new DefaultKinveySyncCallback(latch);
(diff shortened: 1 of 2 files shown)


android-lib/src/androidTest/java/com/kinvey/androidTest/store/DataStoreTest.java
```diff
@@ -1489,6 +1504,190 @@ public void testPagedPullBlocking() throws InterruptedException, IOException {
...
+        assertEquals(5, getCacheSize(StoreType.CACHE));
+        for (int i = 0; i < 5; i++) {
+            query.setLimit(1);
+            query.setSkip(i);
+            pullResults = store.pullBlocking(query).getResult();
+
+            assertNotNull(pullResults);
+            assertTrue(pullResults.size() == 1);
+            assertEquals(5, getCacheSize(StoreType.CACHE));
+        }
```
vinaygahlawat: Would it make sense in these tests to check each entity as it is returned to verify that it is the one that we expect, since we have a sort order?

yuliya-guseva: It makes sense because in this way I check that sorting works correct after getting the cache. If sorting won't work correct user can lose them some data.

PR #126, opened 2017-12-14
feat: MLIBZ-2240 calling "hashCode" on an entity throws an exception if local items were synced with 3.0.7

yuliya-guseva:
#### Description
SDK throws exception after migration from 3.0.7 and below to 3.0.9 and higher after calling `hashcode` method from any item.

#### Changes
Improved migration. The reason for the exception is tables which weren't renamed because they had incorrect names, like book_acl_kmd. Acl table shouldn't have the _kmd table. This tables will be removed after the first treatment to the local collection. 

#### Tests
Instrumented test (check that item has a hashcode), manual test.

### DIFF
--- a/android-lib/src/androidTest/java/com/kinvey/androidTest/store/DataStoreTest.java
+++ b/android-lib/src/androidTest/java/com/kinvey/androidTest/store/DataStoreTest.java
@@ -2441,26 +2441,15 @@ public class DataStoreTest {
     public void testHashCode() throws InterruptedException, IOException {
         DataStore<Person> store = DataStore.collection(Person.COLLECTION, Person.class, StoreType.SYNC, client);
         client.getSyncManager().clear(Person.COLLECTION);
-        cleanBackendDataStore(store);
-        store.syncBlocking(null);
+        clearBackend(store);
         Person person = createPerson(TEST_USERNAME);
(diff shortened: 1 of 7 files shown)


android-lib/src/androidTest/java/com/kinvey/androidTest/store/DataStoreTest.java
```diff
@@ -2437,6 +2437,21 @@ public void testInitializationInternalTable() throws InterruptedException, IOExc
...
+        client.getSyncManager().clear(Person.COLLECTION);
+        clearBackend(store);
+        Person person = createPerson(TEST_USERNAME);
+        DefaultKinveyClientCallback saveCallback = save(store, person);
+        assertNotNull(saveCallback.result);
+        assertNull(saveCallback.error);
+        store.syncBlocking(null);
+        List<Person> personList = store.find();
+        Person person1 = personList.get(0);
+        assertNotNull(person1.hashCode());
```
vinaygahlawat: I think a good addition to this test would be to perform multiple finds in a row on the same object, and verify that the hash code is consistently the same.

yuliya-guseva: Updated

PR #127, opened 2017-12-15
feat: MLIBZ-2241 Sync with delta set caching set to true fails with exception

yuliya-guseva:
#### Description
Sync method throws an exception if Delta Cache is enabled

#### Changes
Changed response parser type to array

#### Tests
Instrumented test (check that sync method doesn't throw exception if Delta Cache is enabled), manual test

### DIFF
--- a/android-lib/src/androidTest/java/com/kinvey/androidTest/store/DataStoreTest.java
+++ b/android-lib/src/androidTest/java/com/kinvey/androidTest/store/DataStoreTest.java
@@ -2437,6 +2437,18 @@ public class DataStoreTest {
         }
     }
 
+    @After
+    public void tearDown() {
+        client.performLockDown();
+        if (client.getKinveyHandlerThread() != null) {
+            try {
+                client.stopKinveyHandlerThread();
+            } catch (Throwable throwable) {
+                throwable.printStackTrace();
+            }
+        }
+    }
+
     @Test
     public void testDeltaCache() throws InterruptedException, IOException {
         client.setUseDeltaCache(true);
@@ -2477,18 +2489,6 @@ public class DataStoreTest {
         DataStore<Person> store = DataStore.collection(Person.COLLECTION, Person.class, StoreType.SYNC, client);
         client.setUseDeltaCache(true);
         assertFalse(store.isDeltaSetCachingEnabled());
-    }
-
-    @After
-    public void tearDown() {
-        client.performLockDown();
-        if (client.getKinveyHandlerThread() != null) {
-            try {
-                client.stopKinveyHandlerThread();
-            } catch (Throwable throwable) {
-                throwable.printStackTrace();
-            }
-        }
     }
 
 }
--- a/android-lib/src/androidTest/java/com/kinvey/androidTest/store/DataStoreTest.java
+++ b/android-lib/src/androidTest/java/com/kinvey/androidTest/store/DataStoreTest.java
@@ -2437,6 +2437,48 @@ public class DataStoreTest {
         }
     }
 
+    @Test
+    public void testDeltaCache() throws InterruptedException, IOException {
+        client.setUseDeltaCache(true);
+        DataStore<Person> store = DataStore.collection(Person.COLLECTION, Person.class, StoreType.SYNC, client);
+        assertTrue(store.isDeltaSetCachingEnabled());
+        client.getSyncManager().clear(Person.COLLECTION);
+        clearBackend(store);
+
+        Person person = new Person();
+        person.setUsername("name");
+        person.set("CustomField", "CustomValue");
+        store.save(person);
+
+        store.syncBlocking(null);
+        Person person2 = new Person();
+        person2.setUsername("changed name 1");
+        person2.set("CustomField", "CustomValueChanged ");
+        store.save(person2);
+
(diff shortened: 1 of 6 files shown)


android-lib/src/androidTest/java/com/kinvey/androidTest/store/DataStoreTest.java
```diff
@@ -2449,4 +2449,46 @@ public void tearDown() {
...
+        assertNotNull(callback.kinveyPullResponse);
+
+        List<Person> personList = store.find();
+        Person person1 = personList.get(0);
+        assertNotNull(person1);
+        assertNotNull(person1.getUsername());
+    }
+
+    @Test
+    public void testDeltaCacheAfterDataStoreInitialization() throws InterruptedException, IOException {
```
tejasranade: This test describes how the SDK behaves today, but I think we should improve the behavior.

I recommend logging the following as a separate enhancement -
The `setUseDeltaCache` flag should be on the datastore instance. It should be set after the datastore is created.

For this ticket, it is ok to leave the test in.

android-lib/src/androidTest/java/com/kinvey/androidTest/store/DataStoreTest.java
```diff
@@ -2449,4 +2449,46 @@ public void tearDown() {
         }
     }
 
+    @Test
+    public void testDeltaCache() throws InterruptedException, IOException {
```
tejasranade: I don't understand how this is testing delta set caching. I think you should add more thorough tests, ideally with mocked network responses. For example 

- Pull a collection with `n` items into a fresh datastore. Then pull it again. The delta set should be of size zero, and cache should have the `n` items in it.
- Pull a collection with `n` items into a fresh datastore. Add or update an item on the backend, then pull again. The delta should be of size 1, and the new cache should reflect the new state of the backend.
... and so on.

iOS has a set of tests along such lines - https://github.com/Kinvey/swift-sdk/blob/f8302e75ffd11435b51625110dff746c38585a99/Kinvey/KinveyTests/DeltaSetCacheTestCase.swift#L634

java-api-core/src/com/kinvey/java/network/NetworkManager.java
```diff
@@ -740,15 +737,14 @@ public Aggregate aggregate(ArrayList<String> fields, AggregateType type, String
 
                 Query query = query().in("_id", chunkItems.toArray((String[])Array.newInstance(String.class, chunkSize)));
                 Get pageGet = new Get(query,
-                        getResponseClass(),
+                        Array.newInstance(myClass,0).getClass(),
```
tejasranade: I looked up the history of this file, and noticed that this line had been incorrect for a long time (2 years). Based on that, we should've had this issue all along. How is it that it was reported as a recent regression? Did something else change that caused the regression?

android-lib/src/androidTest/java/com/kinvey/androidTest/store/DataStoreTest.java
yuliya-guseva: Before fix the `store.syncBlocking(null)` line had an exception, and here it's checking that after the fix SDK doesn't have this exception.
This is the issue described in the ticket.
There's a ticket for Delta Cache tests' improvements: `MLIBZ-2279 Add tests for CRUD operations with delta set caching enabled`. I've added tests for checking that CRUD operation works if Delta Cache is enabled. https://github.com/Kinvey/java-library/pull/128 . 
About the cases with checking the size of items which were got from the server if Delta Cache is enabled: SDK merges these items with items from cache inside DeltaPull (or DeltaGet) requests, these requests have the private method with the additional request for this. 
I can try to add test to this method, I think it will be good to do in this branch https://github.com/Kinvey/java-library/pull/128. 
But I'm not sure that I'll finish it before Friday's release.

android-lib/src/androidTest/java/com/kinvey/androidTest/store/DataStoreTest.java
yuliya-guseva: It makes sense! I'll create a ticket for this.

java-api-core/src/com/kinvey/java/network/NetworkManager.java
yuliya-guseva: I've created `Pull` and `DeltaPull` requests 27-31.10.17. Before SDK used `Get` requests. It was updated in order to add a list of exceptions in the Response. Currently `Pull` and `DeltaPull` requests have `KinveyAbstractReadResponse` which has a `listOfException` field. But `Get` request is used inside `DeltaPull` to get items' ids for Delta Cache logic, and this Get request in `DeltaPull` has some differences in initialization in comparison with `DeltaGet` because `DeltaGet` extends `Get` but `DeltaPull` isn't.

android-lib/src/androidTest/java/com/kinvey/androidTest/store/DataStoreTest.java
tejasranade: Makes sense. I'll review your tests in #128. I understand the tricky part with the private methods - it's not absolutely necessary, but we should add it at some point.
If #128 looks good, we can merge these two open PRs and release.

PR #128, opened 2017-12-27
feat: MLIBZ-2279 Add tests for CRUD operations with delta set caching enabled

yuliya-guseva:
#### Description
Added instrumented tests for CRUD operations (sync and async) with delta set caching enabled. 

#### Changes
Added instrumented tests for delta set caching.

#### Tests
Instrumented

### DIFF
--- a/android-lib/src/androidTest/java/com/kinvey/androidTest/store/DeltaCacheTest.java
+++ b/android-lib/src/androidTest/java/com/kinvey/androidTest/store/DeltaCacheTest.java
@@ -1,0 +1,371 @@
+package com.kinvey.androidTest.store;
+
+import android.content.Context;
+import android.support.test.InstrumentationRegistry;
+import android.support.test.runner.AndroidJUnit4;
+import android.test.RenamingDelegatingContext;
+import android.test.suitebuilder.annotation.SmallTest;
+
+import com.kinvey.android.Client;
+import com.kinvey.android.store.DataStore;
+import com.kinvey.androidTest.TestManager;
+import com.kinvey.androidTest.callback.CustomKinveySyncCallback;
+import com.kinvey.androidTest.callback.DefaultKinveyClientCallback;
+import com.kinvey.androidTest.callback.DefaultKinveyDeleteCallback;
+import com.kinvey.androidTest.model.Person;
+import com.kinvey.java.store.StoreType;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.io.IOException;
+import java.util.List;
+
+import static com.kinvey.androidTest.TestManager.PASSWORD;
+import static com.kinvey.androidTest.TestManager.USERNAME;
+import static junit.framework.Assert.assertEquals;
+import static junit.framework.Assert.assertFalse;
+import static junit.framework.Assert.assertNotNull;
+import static junit.framework.Assert.assertNull;
+import static junit.framework.Assert.assertTrue;
+
+/**
+ * Created by yuliya on 12/27/17.
+ */
+
+@RunWith(AndroidJUnit4.class)
+@SmallTest
+public class DeltaCacheTest {
+
+    private Client client;
+    private TestManager<Person> testManager;
+    private DataStore<Person> store;
+
+    @Before
+    public void setUp() throws InterruptedException, IOException {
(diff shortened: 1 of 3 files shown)


android-lib/src/androidTest/java/com/kinvey/androidTest/store/DeltaCacheTest.java
```diff
@@ -0,0 +1,371 @@
...
+import static junit.framework.Assert.assertNull;
+import static junit.framework.Assert.assertTrue;
+
+/**
+ * Created by yuliya on 12/27/17.
+ */
+
+@RunWith(AndroidJUnit4.class)
+@SmallTest
+public class DeltaCacheTest {
```
tejasranade: Overall, these tests cover a basic set of criteria for CRUD on a single item. 
You should add a few tests on `read`, that cover reading multiple items from the backend. There should be at least one test for `read all` and one for `read by query`.

android-lib/src/androidTest/java/com/kinvey/androidTest/store/DeltaCacheTest.java
```diff
@@ -0,0 +1,461 @@
...
+    }
+
+    @Test
+    public void testReadALotOfItems() throws InterruptedException, IOException {
+        client.setUseDeltaCache(true);
+        store = DataStore.collection(Person.COLLECTION, Person.class, StoreType.SYNC, client);
+        assertTrue(store.isDeltaSetCachingEnabled());
+        testManager.cleanBackend(store, StoreType.SYNC);
+
+        for (int i = 0; i < 10; i++) {
```
tejasranade: I have mentioned this to @vinaygahlawat before as well (for .NET). It will be cleaner to have some common utility functions that prepare data for all tests. Something you should think about as a future improvement.
Something like `createNEntities(int n)`

android-lib/src/androidTest/java/com/kinvey/androidTest/store/DeltaCacheTest.java
```diff
@@ -0,0 +1,461 @@
...
+            Person person = new Person();
+            person.setUsername("DeltaCacheUserName_" + i);
+            Person savedPerson = store.save(person);
+            assertNotNull(savedPerson);
+            assertEquals("DeltaCacheUserName_" + i, savedPerson.getUsername());
+        }
+
+        store.pushBlocking();
+        List<Person> pulledPersons = store.pullBlocking(client.query()).getResult();
```
tejasranade: Before making this first pull, shouldn't you clear the cache? Otherwise, this is also a delta set fetch (since the items you saved are already in the cache).

android-lib/src/androidTest/java/com/kinvey/androidTest/store/DeltaCacheTest.java
```diff
@@ -0,0 +1,461 @@
...
+        assertNotNull(pulledPersons);
+        assertEquals(10, pulledPersons.size());
+
+        Person person;
+        for (int i = 0; i < 10; i++) {
+            person = pulledPersons.get(i);
+            assertEquals("DeltaCacheUserName_" + i, person.getUsername());
+        }
+
+        pulledPersons = store.pullBlocking(client.query()).getResult();
```
tejasranade: One thing you could do is use a separate store of type `NETWORK` to make changes on the backend (e.g. add a new entity, update an existing one). Then your subsequent pull can verify that the addition or update is reflected in the results.

yuliya-guseva: Added

android-lib/src/androidTest/java/com/kinvey/androidTest/store/DeltaCacheTest.java
yuliya-guseva: Added to one of the tests

android-lib/src/androidTest/java/com/kinvey/androidTest/store/DeltaCacheTest.java
yuliya-guseva: Created in TestManager class

android-lib/src/androidTest/java/com/kinvey/androidTest/store/DeltaCacheTest.java
yuliya-guseva: Added

PR #129, opened 2018-01-03
feat: MLIBZ-2144 Implement #clear and #purge by query

yuliya-guseva:
#### Description
Implemented `clear` and `purge` by query.

#### Changes
Added query parameter to `clear` and `purge` methods.

#### Tests
Instrumented

### DIFF
--- a/android-lib/src/main/java/com/kinvey/android/store/DataStore.java
+++ b/android-lib/src/main/java/com/kinvey/android/store/DataStore.java
@@ -122,6 +122,7 @@ public class DataStore<T extends GenericJson> extends BaseDataStore<T> {
     private static final String KEY_DELETE_BY_IDS = "KEY_DELETE_BY_IDS";
 
     private static final String KEY_PURGE = "KEY_PURGE";
+    private static final String KEY_PURGE_BY_QUERY = "KEY_PURGE_BY_QUERY";
 
 
     /*private static final String KEY_GET_BY_ID_WITH_REFERENCES = "KEY_GET_BY_ID_WITH_REFERENCES";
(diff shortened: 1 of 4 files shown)


java-api-core/src/com/kinvey/java/store/BaseDataStore.java
```diff
@@ -417,6 +434,17 @@ public void purge() {
         client.getSyncManager().clear(collection);
     }
 
+    public void purge(Query query) {
+        Preconditions.checkArgument(storeType != StoreType.NETWORK, "InvalidDataStoreType");
+        Preconditions.checkNotNull(client, "client must not be null.");
+        Preconditions.checkArgument(client.isInitialize(), "client must be initialized.");
+        for (T item : cache.get(query)) {
+            if (item.get("_id") != null) {
```
vinaygahlawat: Since item.get("_id") is potentially being called twice here, we should save the value to a local variable and use it twice.

android-lib/src/main/java/com/kinvey/android/store/DataStore.java
```diff
@@ -681,7 +683,7 @@ public void pull(KinveyPullCallback<T> callback) {
     }
 
     /**
-     * Asynchronous request to clear cache from a collection of entity and pull all collection.
+     * Asynchronous request to clear all the pending requests from the sync storage
      * <p>
      * Creates an asynchronous request to clear cache from a collection of entity and pull all collection.
```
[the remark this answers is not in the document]
yuliya-guseva: Updated

java-api-core/src/com/kinvey/java/store/BaseDataStore.java
yuliya-guseva: Improved

PR #130, opened 2018-01-08
feat: MLIBZ-2283 Fix setting request timeout

yuliya-guseva:
#### Description
`Client.setRequestTimeout` method didn't affect to request timeout. It was fixed.

#### Changes
Call `setRuestTimeout` was added  to `Client.Builder.build` method.
Default request timeout was updated from 20s to 60s.

#### Tests
Instrumented, manual.

### DIFF
--- a/android-lib/src/androidTest/java/com/kinvey/androidTest/request/RequestTest.java
+++ b/android-lib/src/androidTest/java/com/kinvey/androidTest/request/RequestTest.java
@@ -118,6 +118,7 @@ public class RequestTest {
 
         DefaultKinveyPushCallback pushCallback = testManager.push(store);
         assertNotNull(pushCallback);
+        assertNull(pushCallback.getResult());
         assertNotNull(pushCallback.getError());
         assertEquals("SocketTimeoutException", pushCallback.getError().getClass().getSimpleName());
     }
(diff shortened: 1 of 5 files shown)


android-lib/src/androidTest/java/com/kinvey/androidTest/request/RequestTest.java
```diff
@@ -0,0 +1,151 @@
...
+        assertNotNull(callback);
+        assertNotNull(callback.getResult());
+        assertNull(callback.getError());
+        assertNotNull(callback.getResult().getUsername());
+        assertEquals(TEST_USERNAME, callback.getResult().getUsername());
+
+        DefaultKinveyPushCallback pushCallback = testManager.push(store);
+        assertNotNull(pushCallback);
+        assertNotNull(pushCallback.getError());
+        assertEquals("SocketTimeoutException", pushCallback.getError().getClass().getSimpleName());
```
vinaygahlawat: Should we assertNull(pushCallback.getResult()) here?

yuliya-guseva: In general if `pushCallback` has error it shouldn't have any result in callback. But I think it's good to have a double check on this logic. `assertNull(pushCallback.getResult())` was added.

PR #137, opened 2018-01-25
feat: MLIBZ-2318 Add sorting for tests with items order checking

yuliya-guseva:
#### Description
For correct work `skip` parameter a user have to add sorting to the query. 
If autopagination is enabled, sorting should work correct without any additinal action from the user side.

#### Changes
Added sorting for tests with items order checking. Also sorting was added in autopagination logic. 
In autopagination case SDK didn't lose user data without sorting, but had not correct items order as well.

#### Tests
Instrumented

### DIFF
--- a/android-lib/src/androidTest/java/com/kinvey/androidTest/store/DeltaCacheTest.java
+++ b/android-lib/src/androidTest/java/com/kinvey/androidTest/store/DeltaCacheTest.java
@@ -14,6 +14,8 @@ import com.kinvey.androidTest.callback.CustomKinveySyncCallback;
 import com.kinvey.androidTest.callback.DefaultKinveyClientCallback;
 import com.kinvey.androidTest.callback.DefaultKinveyDeleteCallback;
 import com.kinvey.androidTest.model.Person;
+import com.kinvey.java.Query;
+import com.kinvey.java.query.AbstractQuery;
 import com.kinvey.java.store.StoreType;
 
 import org.junit.After;
@@ -149,7 +151,8 @@ public class DeltaCacheTest {
         testManager.createPersons(store, TEN_ITEMS);
 
         store.pushBlocking();
-        List<Person> pulledPersons = store.pullBlocking(client.query()).getResult();
+        Query query = client.query().addSort("_kmd", AbstractQuery.SortOrder.ASC);
+        List<Person> pulledPersons = store.pullBlocking(query).getResult();
         assertNotNull(pulledPersons);
         assertEquals(TEN_ITEMS, pulledPersons.size());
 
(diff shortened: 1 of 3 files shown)


android-lib/src/androidTest/java/com/kinvey/androidTest/store/DeltaCacheTest.java
```diff
@@ -568,12 +572,12 @@ public void testUpdateWithTwoStorageTypes() throws InterruptedException, IOExcep
...
         assertEquals(TEST_USERNAME + 100, saveCallback.getResult().getUsername());
 
-        CustomKinveyPullCallback<Person> pullCallback = testManager.pullCustom(store, client.query());
+        CustomKinveyPullCallback<Person> pullCallback = testManager.pullCustom(store, client.query().addSort("_kmd", AbstractQuery.SortOrder.ASC));
         assertNotNull(pullCallback);
         assertNotNull(pullCallback.getResult());
         assertNull(pullCallback.getError());
         assertEquals(TEN_ITEMS, pullCallback.getResult().getResult().size());
-        assertEquals(TEST_USERNAME + 100, pullCallback.getResult().getResult().get(0).getUsername());
+        assertEquals(TEST_USERNAME + 100, pullCallback.getResult().getResult().get(9).getUsername());
```
vinaygahlawat: Why did the index change here from **_0_** to **_9_**?

android-lib/src/androidTest/java/com/kinvey/androidTest/store/DeltaCacheTest.java
```diff
@@ -149,7 +151,8 @@ public void testReadALotOfItems() throws InterruptedException, IOException {
         testManager.createPersons(store, TEN_ITEMS);
 
         store.pushBlocking();
-        List<Person> pulledPersons = store.pullBlocking(client.query()).getResult();
+        Query query = client.query().addSort("_kmd", AbstractQuery.SortOrder.ASC);
```
vinaygahlawat: Should the sorting here be done by `_kmd.ect` rather than just `_kmd`?

yuliya-guseva: In the test doesn't matter what we set in sort parameters, there is important that query has any sorting.

android-lib/src/androidTest/java/com/kinvey/androidTest/store/DeltaCacheTest.java
yuliya-guseva: I check the 9th (last item) because if we set sorting only by `_kmd` query will be sorted by `_kmd.lmt`' because `_kmd` object has `lmt` parameter before `ect`, example: `{"lmt":"2018-01-12T11:00:01.122Z","ect":"2018-01-12T11:00:01.122Z"}`.

PR #138, opened 2018-02-02
feat: MLIBZ-2339 Add support for null nested objects

yuliya-guseva:
#### Description
User could not set null to the nested object and save it. SDK has ignored that nested object was changed to null.

#### Changes
SDK has ignored fields with null objects, this logic was updated. Currently If an object has a field with a null object it will be saved to the realm object in the field with a null value.

#### Tests
Instrumented

### DIFF
--- a/java-api-core/src/com/kinvey/java/model/KinveyMetaData.java
+++ b/java-api-core/src/com/kinvey/java/model/KinveyMetaData.java
@@ -22,6 +22,7 @@ import com.kinvey.java.AbstractClient;
 
 import java.util.ArrayList;
 import java.util.Calendar;
+import java.util.Locale;
 import java.util.Map;
 import java.util.TimeZone;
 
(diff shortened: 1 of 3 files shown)


java-api-core/src/com/kinvey/java/model/KinveyMetaData.java
```diff
@@ -61,13 +62,13 @@ public static KinveyMetaData fromMap(Map kmd) {
         KinveyMetaData metaData = new KinveyMetaData();
         if (kmd != null) {
             if (!kmd.containsKey(LMT) || kmd.get(LMT) == null) {
-                metaData.put(LMT, String.format("%tFT%<tTZ",
+                metaData.put(LMT, String.format(Locale.US, "%tFT%<tTZ",
```
vinaygahlawat: @yuliya-guseva What is the reason for putting `Locale.US` in as a formatter? And does this affect anything from a migration perspective?

yuliya-guseva: I've added this improvement, because of I've noticed that SDK uses Locale.US to converting headers in the requests to lower case ` toLowerCase(Locale.US)`. Also I see that method `String.format` has the recommendation in java doc:
```* Returns a localized formatted string, using the supplied format and arguments, * using the user's default locale. * * <p>If you're formatting a string other than for human * consumption, you should use the {@code format(Locale, String, Object...)} * overload and supply {@code Locale.US}. See * "<a href="../util/Locale.html#default_locale">Be wary of the default locale</a>".```
I've checked migration from 3.0.13 to the build with this fix, all is OK. I think this hasn't to cause any problems.
But if You think that putting Local.US in can create some issues, we can postpone it until better times.

vinaygahlawat: If the migration tests pass, it should be fine to put in the local change now.

PR #140, opened 2018-02-09
Feature/mlibz 2171 logout does not invalidate access token

vinaygahlawat:
#### Description
Logout call was only acting locally, not making `_logout` request to backend.

#### Changes
Making the logout request now as part of hard logout.

#### Tests
Instrumented, manual


vinaygahlawat:
Merging now, although there may be some issues with Travis integration tests. All tests are passing locally.

PR #141, opened 2018-02-14
feat: MLIBZ-2328 Add support for long names in fields with List type

yuliya-guseva:
#### Description
If ArrayList<String> field with the long name (more than 7 chars) is used in a data model class, SDK had the exception like "Class name is too long".

#### Changes
SDK creates the additional table with the short name for keeping Lists. Before this logic worked only for embedded objects.

#### Tests
Instrumented

### DIFF
--- a/android-lib/src/main/java/com/kinvey/android/cache/ClassHash.java
+++ b/android-lib/src/main/java/com/kinvey/android/cache/ClassHash.java
@@ -382,7 +382,7 @@ public abstract class ClassHash {
                 } else {
                     for (Class c : ALLOWED) {
                         if (underlying.equals(c)) {
-                            RealmObjectSchema innerScheme = realm.getSchema().create(shortName + "_" + fieldInfo.getName());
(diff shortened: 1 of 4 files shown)


android-lib/src/main/java/com/kinvey/android/cache/ClassHash.java
```diff
@@ -695,7 +695,7 @@ static int deleteClassData(String collection, DynamicRealm realm, Class<? extend
             }
             if (fieldInfo.getType().isArray() || Collection.class.isAssignableFrom(fieldInfo.getType())) {
                 Class underlying = getUnderlying(f);
-                if (underlying != null && GenericJson.class.isAssignableFrom(underlying)) {
+                if (underlying != null) {
```
vinaygahlawat: Why is the check for GenericJson assignment removed for this fix?

yuliya-guseva: Because realm keeps a list of any items like a small table so SDK should have one rule for deleting the list of any types.

vinaygahlawat: 👍

PR #142, opened 2018-02-15
feat: MLIBZ-2360 Integrate PubNub library

yuliya-guseva:
#### Description
Investigate how to integrate PubNub to the SDK. Integrate PubNub library.

#### Changes
Added PubNub library as dependency in the build.gradle.


yuliya-guseva:
Ok

PR #143, opened 2018-02-19
feat: MLIBZ-2361 Implement live service router to integrate with PubNub library

yuliya-guseva:
#### Description
Implement live service router to integrate with pub nub library

#### Changes
Added singleton RealtimeRouter class with initialize and unInitialize methods. 

#### Tests
Will be tested in the MLIBZ-2362 branch with register/unregister methods

### DIFF
--- a/java-api-core/src/com/kinvey/java/store/RealtimeRouter.java
+++ b/java-api-core/src/com/kinvey/java/store/RealtimeRouter.java
@@ -27,7 +27,7 @@ class RealtimeRouter {
 
     }
 
-    protected static RealtimeRouter getInstance() {
+    static RealtimeRouter getInstance() {
         synchronized (lock) {
             if (realtimeRouter == null) {
                 realtimeRouter = new RealtimeRouter();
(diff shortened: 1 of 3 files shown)


java-api-core/src/com/kinvey/java/store/RealtimeRouter.java
```diff
@@ -0,0 +1,99 @@
...
+            if (realtimeRouter == null) {
+                realtimeRouter = new RealtimeRouter();
+            }
+            return realtimeRouter;
+        }
+    }
+
+    void initialize(String channelGroup, String publishKey, String subscribeKey, String authKey, AbstractClient client) {
+        synchronized (lock) {
+            if (!isInitialized()) {
```
vinaygahlawat: Do we need to acquire this lock every time initialized is called, even if the `RealtimeRouter` has already been initialized? We may be able to make this more efficient by doing [double-check locking](https://en.wikipedia.org/wiki/Double-checked_locking).

java-api-core/src/com/kinvey/java/store/RealtimeRouter.java
```diff
@@ -0,0 +1,99 @@
...
+                pubnubClient.addListener(subscribeCallback);
+                pubnubClient.subscribe().channelGroups(Collections.singletonList(channelGroup)).execute();
+            }
+        }
+
+    }
+
+    void unInitialize() {
+        synchronized (lock) {
+            if (isInitialized()) {
```
vinaygahlawat: Same question here about locking as in the `initialize()` method.

java-api-core/src/com/kinvey/java/store/RealtimeRouter.java
yuliya-guseva: I agree. Updated.

java-api-core/src/com/kinvey/java/store/RealtimeRouter.java
yuliya-guseva: Updated.

yuliya-guseva:
Changed `RealtimeRouter` to `LiveServiceRouter`.

PR #144, opened 2018-02-19
feat: MLIBZ-2362 Implement Register/Unregister methods

yuliya-guseva:
#### Description
Implement Register/Unregister methods

#### Changes
Added register/unregister methods to User.

#### Tests
Instrumented

### DIFF
--- a/android-lib/src/main/java/com/kinvey/android/Client.java
+++ b/android-lib/src/main/java/com/kinvey/android/Client.java
@@ -1064,5 +1064,10 @@ public class Client<T extends User> extends AbstractClient<T> {
             kinveyHandlerThread.interrupt();
         }
     }
+
+    @Override
+    public String getDeviceId() {
+        return new UdidFactory(context).getDeviceUuid().toString();
+    }
 }
 
--- a/android-lib/src/main/java/com/kinvey/android/Client.java
+++ b/android-lib/src/main/java/com/kinvey/android/Client.java
@@ -1067,7 +1067,7 @@ public class Client<T extends User> extends AbstractClient<T> {
 
     @Override
     public String getDeviceId() {
-        return new UdidFactory(context).getDeviceUuid().toString();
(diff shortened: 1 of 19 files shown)


android-lib/src/main/java/com/kinvey/android/Client.java
```diff
@@ -1064,5 +1064,10 @@ public void stopKinveyHandlerThread() {
             kinveyHandlerThread.interrupt();
         }
     }
+
+    @Override
+    public String getDeviceId() {
+        return new UdidFactory(context).getDeviceUuid().toString();
```
vinaygahlawat: If we can, we should fix the misspelling of `UdidFactory`.

yuliya-guseva: This is internal class, we can. Updated.

java-api-core/src/com/kinvey/java/store/BaseUserStore.java
```diff
@@ -143,6 +145,42 @@ public static boolean exists( String username, AbstractClient client) throws IOE
...
+
+    /**
+     * Unregister the active user from realtime messaging.
+     * @throws IOException
+     */
+    public static void unRegisterRealtime() throws IOException {
+        if (AbstractClient.sharedInstance().getActiveUser() == null) {
+            throw new KinveyException("User object has to be the active user in order to register for realtime messages");
+        }
+        RealtimeRouter.getInstance().unInitialize();
```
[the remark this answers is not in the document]
yuliya-guseva: If the RealtimeRouter is not initialized, router won't initialize him to uninitialize. The Router's methods has initialize checking before executing. But I've added additianal isInitialize checking to BaseUserStore to don't make a unnessesary request to the server.

yuliya-guseva:
Currently we have one small test for register/unregister. DeviceId already is kept in the SharedPreferences. 
Added small device uuid test.

PR #148, opened 2018-03-03
feat: MLIBZ-2185 Clear does not clear metadata for data synced in previous sdk versions

yuliya-guseva:
#### Description
`DataStore#clear` does not clear metadata for data synced in previous SDK versions.

#### Changes
Added new method `clearCollection` to CacheManager. The method deletes the collection and creates its new version. Current method `clear` just deletes realm file, this logic was saved. Logic for `DataStore.clear` was updated from `CacheManager#clear` to `CacheManager#clearCollection`.

#### Tests
Instrumented, Manual. Manual testing is necessary in this case because we have to update library version from 3.0.5 to latest version, it's impossible to check in the tests. Anyway, the small test for checking was added, the test checks that `CacheManager#clearCollection` doesn't crash.

### DIFF
--- a/android-lib/src/main/java/com/kinvey/android/cache/RealmCacheManager.java
+++ b/android-lib/src/main/java/com/kinvey/android/cache/RealmCacheManager.java
@@ -211,6 +211,51 @@ public class RealmCacheManager implements ICacheManager {
         }
     }
 
+    @Override
+    public <T extends GenericJson> void clearCollection(String collection, Class<T> collectionItemClass, Long ttl) {
+        synchronized (LOCK) {
+            DynamicRealm realm = getDynamicRealm();
+            try {
+                realm.beginTransaction();
+                List<String> schemas = new ArrayList<>();
+                schemas.add(TableNameManager.getShortName(collection, realm));
+                schemas.addAll(getEmbeddedObjectSchemas(collection, realm));
+                removeSchemas(schemas, realm);
+                String cacheKey = getClientHash() + File.separator + collection;
+                RealmCache<T> cache = (RealmCache<T>) mCacheMap.get(cacheKey);
+                if (cache != null) {
+                    cache.createRealmTable(realm);
+                }
+            } finally {
+                realm.commitTransaction();
+            }
+        }
+    }
+
+    private List<String> getEmbeddedObjectSchemas(String tableName, DynamicRealm realm) {
+        RealmSchema currentSchema = realm.getSchema();
+        String originalName;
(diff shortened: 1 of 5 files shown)


android-lib/src/main/java/com/kinvey/android/cache/RealmCacheManager.java
```diff
@@ -195,6 +195,7 @@ public RealmCacheManager(String prefix, Client client) {
      */
     private void removeSchemas(List<String> tableNames, DynamicRealm realm) {
         RealmSchema realmSchema = realm.getSchema();
+        System.out.println(tableNames);
```
vinaygahlawat: Please remove any debugging statements if not necessary.

android-lib/src/main/java/com/kinvey/android/cache/RealmCacheManager.java
```diff
@@ -211,6 +212,52 @@ public void clear() {
...
+        Set<RealmObjectSchema> schemas = currentSchema.getAll();
+        for (RealmObjectSchema schema : schemas) {
+            className = schema.getClassName();
+            if (className.equals(TableNameManager.getShortName(tableName, realm))) {
+                for (RealmObjectSchema subClassSchema : schemas) {
+                    originalName = TableNameManager.getOriginalName(subClassSchema.getClassName(), realm);
+                    if (originalName == null) {
+                        continue;
+                    }
+                    if (originalName.startsWith(className + "_")) {
```
vinaygahlawat: Should use the underscore constant here if possible.

yuliya-guseva: I agree. Constants.UNDERSCORE was added to another branch (MLIBZ-2392). I'll update it after merging MLIBZ-2392 branch to indev.

yuliya-guseva:
`CacheManager#clearCollection` is used in `DataStore#clear`, because the previous logic of `DataStore#clear` deleted realm file with all tables instead of clearing only selected collection. But this logic when SDK have to delete realm file is used in the `Client#performLockdown` so I just added the new method in the RealmCachemanager(clearCollection) and updated `DataStore#clear`. 
`Client#performLockdown` also is used in the `Logout` request and `Delete(hard)` user, probably  should we use `CacheManager#clearCollection` in this case as well and not delete all data in the realm?

vinaygahlawat:
I think that for the `Logout` case, we still want to delete all the data in Realm.

PR #152, opened 2018-03-26
feat: MLIBZ-2333 Add method for setting instanceID

yuliya-guseva:
#### Description
Add method for setting instance

#### Changes
Added method for setting instanceID to Client.Builder. 

#### Tests
Instrumented

### DIFF
--- a/android-lib/src/main/java/com/kinvey/android/Client.java
+++ b/android-lib/src/main/java/com/kinvey/android/Client.java
@@ -779,6 +779,14 @@ public class Client<T extends User> extends AbstractClient<T> {
         }
 
         /**
+         * @param instanceID
+         */
+        public Builder setInstanceID(String instanceID) {
+            super.setInstanceID(instanceID);
+            return this;
+        }
+
+        /**
          * @return an instantiated Kinvey Android Client,
          * which contains factory methods for accessing various functionality.
          */
--- a/android-lib/src/main/java/com/kinvey/android/Client.java
+++ b/android-lib/src/main/java/com/kinvey/android/Client.java
@@ -30,6 +30,7 @@ import com.google.api.client.json.GenericJson;
 import com.google.api.client.json.JsonFactory;
 import com.google.api.client.json.JsonObjectParser;
 import com.google.common.base.Preconditions;
+import com.google.common.base.Strings;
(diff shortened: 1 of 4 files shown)


android-lib/src/main/java/com/kinvey/android/Client.java
```diff
@@ -814,7 +824,9 @@ public Client build(){
             if(this.MICBaseURL != null){
                client.setMICHostName(this.MICBaseURL);
             }
-
+            if (!Strings.isNullOrEmpty(this.instanceID)) {
+                client.setMICHostName(Constants.PROTOCOL_HTTPS + instanceID + Constants.HYPHEN + Constants.HOSTNAME_AUTH);
+            }
```
vinaygahlawat: Will this construct a URL that does not end with a "/"? If so, is this consistent with how the default URL is formatted?

java-api-core/src/com/kinvey/java/AbstractClient.java
```diff
@@ -463,6 +464,12 @@ public Builder setRequestTimeout(int requestTimeout) {
             return this;
         }
 
+        public Builder setInstanceID(String instanceID) {
+            this.instanceID = instanceID;
+            setBaseUrl(Constants.PROTOCOL_HTTPS + instanceID + Constants.HYPHEN + Constants.HOSTNAME_API);
+            return this;
```
vinaygahlawat: Same question as above regarding URL format.

android-lib/src/main/java/com/kinvey/android/Client.java
yuliya-guseva: The SDK has logic for adding  "/" to the end of url. So all URLs will have "/" at the end.
https://github.com/Kinvey/java-library/blob/7c8e85107593870941a39c249e93464bc6238ff3/java-api-core/src/com/kinvey/java/AbstractClient.java#L115
`this.MICHostName = MICHostName.endsWith("/") ? MICHostName : MICHostName + "/"`

java-api-core/src/com/kinvey/java/AbstractClient.java
yuliya-guseva: `setBaseUrl` also has logic for adding "/" to the end of url.

PR #158, opened 2018-03-30
feat: MLIBZ-2103 Improve SaveListRequest for StoreType.CACHE

yuliya-guseva:
#### Description
Improved SaveListRequest performance for StoreType.CACHE

#### Changes
Implemented bulk saving pushed items to the cache

#### Tests
Instrumented

### DIFF
--- a/android-lib/src/androidTest/java/com/kinvey/androidTest/store/DataStoreTest.java
+++ b/android-lib/src/androidTest/java/com/kinvey/androidTest/store/DataStoreTest.java
@@ -2197,6 +2197,8 @@ public class DataStoreTest {
         CustomKinveyListCallback<Person> saveCallback = testManager.saveCustomList(store, persons);
         assertNotNull(saveCallback.getResult());
         assertEquals(3, saveCallback.getResult().size());
+        List<Person> cachedItems = client.getCacheManager().getCache(Person.COLLECTION, Person.class, StoreType.CACHE.ttl).get();
+        assertEquals(3, cachedItems.size());
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


java-api-core/src/com/kinvey/java/store/requests/data/save/SaveListRequest.java
```diff
@@ -51,16 +53,42 @@ public SaveListRequest(ICache<T> cache, NetworkManager<T> networkManager, WriteP
...
+                    try {
+                        ret.add(networkManager.saveBlocking(object).execute());
+                    } catch (IOException e) {
+                        syncManager.enqueueRequest(networkManager.getCollectionName(),
+                                networkManager, RequestMethod.SAVE, (String) object.get(Constants._ID));
+                        notPushedObjects.add(object);
+                        exception = e;
+                    }
+                }
+                cache.save(ret);
```
vinaygahlawat: @yuliya-guseva In the LOCAL_THEN_NETWORK case, is the local cache save happening after the network request? If so, it may make more sense to first save locally to cache before attempting the network operation.

android-lib/src/androidTest/java/com/kinvey/androidTest/store/DataStoreTest.java
```diff
@@ -2183,6 +2183,22 @@ public void testSaveList() throws InterruptedException {
...
+        DataStore<Person> store = DataStore.collection(Person.COLLECTION, Person.class, StoreType.CACHE, client);
+        ArrayList<Person> persons = new ArrayList<>();
+        persons.add(new Person(TEST_USERNAME));
+        persons.add(new Person(TEST_USERNAME_2));
+        persons.add(new Person(TEST_TEMP_USERNAME));
+
+        CustomKinveyListCallback<Person> saveCallback = testManager.saveCustomList(store, persons);
+        assertNotNull(saveCallback.getResult());
+        assertEquals(3, saveCallback.getResult().size());
+    }
```
vinaygahlawat: Can an assert be added to check that the size of the cache matches the expected result?

PR #161, opened 2018-04-06
feat: MLIBZ-2387 Update auto pagination to meet design spec

yuliya-guseva:
#### Description
`KinveyPullResponse` has to return count of pulled items instead of list of items. 

#### Changes
The PR has breaking changes:
Users have to: 
1. Remove old import for `KinveyPullResponse`
`import com.kinvey.android.sync.KinveyPullResponse;`
and add a new one `import com.kinvey.java.model.KinveyPullResponse;`.
2. Remove GenericType from `KinveyPullResponse<Book> `, `KinveySyncCallback<Book>`, `KinveyPullCallback<Book>`, it should look like `KinveyPullResponse`, `KinveySyncCallback`, `KinveyPullCallback`.
The method  `public List<T> getResult()` was changed to `public int getCount()` in the `KinveyPullResponse` class.

#### Tests
Instrumented

### DIFF
--- a/java-api-core/src/com/kinvey/java/model/KinveyPullResponse.java
+++ b/java-api-core/src/com/kinvey/java/model/KinveyPullResponse.java
@@ -1,0 +1,15 @@
+package com.kinvey.java.model;
+
+
+public class KinveyPullResponse extends KinveyAbstractResponse {
+
+    private int count;
+
+    public int getCount() {
+        return count;
+    }
+
+    public void setCount(int count) {
+        this.count = count;
+    }
+}
--- a/java-api-core/src/com/kinvey/java/store/BaseDataStore.java
+++ b/java-api-core/src/com/kinvey/java/store/BaseDataStore.java
@@ -394,7 +394,7 @@ public class BaseDataStore<T extends GenericJson> {
 
         KinveyPullResponse response = new KinveyPullResponse();
         query = query == null ? client.query() : query;
-        KinveyAbstractReadResponse<T> pullResponse;
+        KinveyAbstractReadResponse<T> readResponse  ;
         if (isAutoPaginationEnabled()) {
             if (query.getSortString() == null || query.getSortString().isEmpty()) {
                 query.addSort(Constants._ID, AbstractQuery.SortOrder.ASC);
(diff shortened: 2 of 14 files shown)


java-api-core/src/com/kinvey/java/store/BaseDataStore.java
```diff
@@ -385,42 +386,44 @@ public void pushBlocking() throws IOException {
...
         Preconditions.checkArgument(storeType != StoreType.NETWORK, "InvalidDataStoreType");
         Preconditions.checkNotNull(client, "client must not be null.");
         Preconditions.checkArgument(client.isInitialize(), "client must be initialized.");
         Preconditions.checkArgument(client.getSyncManager().getCount(getCollectionName()) == 0, "InvalidOperation. You must push all pending sync items before new data is pulled. Call push() on the data store instance to push pending items, or purge() to remove them.");
 
-        KinveyAbstractReadResponse<T> response = new KinveyAbstractReadResponse<T>();
+        KinveyPullResponse response = new KinveyPullResponse();
         query = query == null ? client.query() : query;
-
+        KinveyAbstractReadResponse<T> pullResponse;
```
vinaygahlawat: Consider renaming this variable to `readResponse`, and changing the name of the `KinveyPullResponse` variable to `pullResponse`.

java-api-core/src/com/kinvey/java/store/BaseDataStore.java
```diff
@@ -385,42 +386,44 @@ public void pushBlocking() throws IOException {
...
+            int pulledItemCount = 0;
             do {
                 query.setSkip(skipCount).setLimit(pageSize);
                 pullResponse = networkManager.pullBlocking(query, cache, isDeltaSetCachingEnabled()).execute();
-                networkData.addAll(pullResponse.getResult());
+                pulledItemCount += pullResponse.getResult().size();
                 exceptions.addAll(pullResponse.getListOfExceptions());
                 cache.delete(query);
-                cache.save(networkData);
+                cache.save(pullResponse.getResult());
```
vinaygahlawat: Is there a way to get `pullItemCount` from the `cache.save()` call, rather than calling `pullResponse.getResult()` twice?

java-api-core/src/com/kinvey/java/store/BaseDataStore.java
```diff
@@ -385,42 +386,44 @@ public void pushBlocking() throws IOException {
...
+            response.setCount(pulledItemCount);
             response.setListOfExceptions(exceptions);
         } else {
-            response = networkManager.pullBlocking(query, cache, isDeltaSetCachingEnabled()).execute();
+            pullResponse = networkManager.pullBlocking(query, cache, isDeltaSetCachingEnabled()).execute();
             cache.delete(query);
-            cache.save(response.getResult());
+            cache.save(pullResponse.getResult());
+            response.setCount(pullResponse.getResult().size());
+            response.setListOfExceptions(pullResponse.getListOfExceptions());
```
vinaygahlawat: Similar question as above, about how to get the pull count from the `cache.save()` call.

java-api-core/src/com/kinvey/java/model/KinveyPullResponse.java
```diff
@@ -0,0 +1,15 @@
+package com.kinvey.java.model;
+
+
+public class KinveyPullResponse extends KinveyAbstractResponse {
```
vinaygahlawat: Out of curiosity, what is the reason for moving the pull response out of `android` and into `java-api-core`?

java-api-core/src/com/kinvey/java/store/BaseDataStore.java
yuliya-guseva: Updated.

java-api-core/src/com/kinvey/java/store/BaseDataStore.java
yuliya-guseva: Yes, `cache.save` returns a list, so we can use size of this list to get saved item count. Updated.

java-api-core/src/com/kinvey/java/store/BaseDataStore.java
yuliya-guseva: The same as in the previous comment. `cache.save` returns a list, we will use size of this list to get saved item count. Updated.

java-api-core/src/com/kinvey/java/model/KinveyPullResponse.java
yuliya-guseva: `pull` from `BaseDataStore` has the same response as `pull` from the DataStore, so I think It's logical to have one response class for both methods.

PR #164, opened 2018-04-26
feat: MLIBZ-2388 Optimize auto-pagination performance

yuliya-guseva:
#### Description
Optimize auto-pagination performance

#### Changes
Pull requests with auto-pagination works concurrently. This logic is similar to push concurrent requests. 

#### Tests
Instrumented


android-lib/src/main/java/com/kinvey/android/store/DataStore.java
```diff
@@ -737,39 +740,41 @@ public void pull(KinveyPullCallback callback) {
      * @param callback KinveyPullCallback
      */
     public void pull(Query query, int pageSize, KinveyPullCallback callback) {
-        Preconditions.checkArgument(pageSize > 0, "pageSize must be more than 0");
+        Preconditions.checkArgument(pageSize >= MIN_PAGE_SIZE, "pageSize mustn't be less than " + MIN_PAGE_SIZE);
```
vinaygahlawat: Please add tests that check that the preconditions correctly work.

yuliya-guseva: Added

PR #167, opened 2018-05-11
feat: MLIBZ-2447 Extending user class with properties of different class cause an exception

yuliya-guseva:
#### Description
Extending user class with properties of different class cause an exception

#### Changes
Updated auth response parsing. Now the SDK gets user class as a model for response parsing. Previously the model for response parsing had only `id` and `_kmd` fields, other primitive fields were got as fields from GenericJson and it didn't work for GenericJson fields in the user model class.

#### Tests
Instrumented

### DIFF
--- a/android-lib/src/androidTest/java/com/kinvey/androidTest/model/TestUser.java
+++ b/android-lib/src/androidTest/java/com/kinvey/androidTest/model/TestUser.java
@@ -8,9 +8,8 @@ public class TestUser extends User {
     @Key("companyName")
     private String companyName;
 
-    public TestUser() {
-
-    }
+    @Key
+    private InternalUserEntity internalUserEntity;
 
     public String getCompanyName() {
         return companyName;
@@ -18,5 +17,13 @@ public class TestUser extends User {
 
     public void setCompanyName(String companyName) {
         this.companyName = companyName;
+    }
+
+    public InternalUserEntity getInternalUserEntity() {
+        return internalUserEntity;
+    }
+
+    public void setInternalUserEntity(InternalUserEntity internalUserEntity) {
+        this.internalUserEntity = internalUserEntity;
     }
 }
--- a/android-lib/src/main/java/com/kinvey/android/Client.java
+++ b/android-lib/src/main/java/com/kinvey/android/Client.java
@@ -431,7 +431,7 @@ public class Client<T extends User> extends AbstractClient<T> {
      * <p/>
      * This Builder class is not thread-safe.
      */
-    public static class Builder extends AbstractClient.Builder {
+    public static class Builder<T extends User> extends AbstractClient.Builder {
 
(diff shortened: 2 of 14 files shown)


android-lib/src/main/java/com/kinvey/android/Client.java
```diff
@@ -431,7 +431,7 @@ public Context getContext(){
      * <p/>
      * This Builder class is not thread-safe.
      */
-    public static class Builder extends AbstractClient.Builder {
+    public static class Builder<T extends User> extends AbstractClient.Builder {
```
vinaygahlawat: Will this be a breaking change to the client builder?

java-api-core/src/com/kinvey/java/AbstractClient.java
```diff
@@ -560,15 +560,15 @@ public SyncManager getSyncManager(){
         return new SyncManager(getSyncCacheManager());
     }
 
-    public Class getUserClass(){
-        return this.userModelClass;
+    public Class<T> getUserClass(){
+        return this.userModelClass != null ? this.userModelClass : BaseUser.class;
```
vinaygahlawat: Why have this logic here instead of defaulting `userModelClass` to `BaseUser.class`?

android-lib/src/androidTest/java/com/kinvey/androidTest/model/TestUser.java
```diff
@@ -8,9 +8,8 @@
     @Key("companyName")
     private String companyName;
 
-    public TestUser() {
-
-    }
+    @Key
+    private InternalUserEntity internalUserEntity;
```
vinaygahlawat: Do we have a test case where the `User` class is extended?

yuliya-guseva: Yes, we have. Previous custom user tests were updated:
https://github.com/Kinvey/java-library/pull/167/files#diff-24be4dde17572f9389b3bbab418535abR378
and
https://github.com/Kinvey/java-library/pull/167/files#diff-24be4dde17572f9389b3bbab418535abR1131

java-api-core/src/com/kinvey/java/AbstractClient.java
yuliya-guseva: It was done because `getUserClass` must return `Class<T>` type, but `BaseUser.class` should be cast to `Class<T>` before. I think I can return old implementation for `getUserClass` and add casting `userModelClass = (Class<T>) BaseUser.class`. It will be more expectable for users. Updated.

android-lib/src/main/java/com/kinvey/android/Client.java
yuliya-guseva: No, I won't be breaking change.

PR #169, opened 2018-05-23
[WIP] MLIBZ-2513 - Locally created entity being pushed as a PUT rather than a POST.

vinaygahlawat:
#### Description
Locally-created entities, unless they have explicitly had an ID set on it, should have the ID removed prior to pushing to the backend, and this should be done via a POST request. Currently, the ID field is being populated by a GUID from Realm, which is causing the push to be done via a PUT request. This is unexpected, and although the Kinvey backend is more forgiving for the PUT vs POST, things such as Flex services may not handle this.

#### Changes
Detect when there is no ID or when the entity has a Realm-generated ID, and remove it prior to performing the push operation. Ensure that this request is sent as a POST. Properly delete the entity in Realm when the response is successful, and add the entity returned, which includes the permanent ID, into Realm.

#### Tests
Instrumented


android-lib/src/main/java/com/kinvey/android/async/CallableAsyncPushRequestHelper.java
```diff
@@ -18,21 +20,34 @@
...
     }
 
     @Override
     public Long call() throws Exception {
         try {
-            manager.executeRequest(client, syncRequest);
+            if (syncRequest.getHttpVerb() == SyncRequest.HttpVerb.POST) {
+                String tempID = syncRequest.getEntityID().id;
+                GenericJson result = manager.executeRequest(client, syncRequest);
+                ICache<T> cache = client.getCacheManager().getCache(syncRequest.getCollectionName(), this.storeItemType, Long.MAX_VALUE);
```
yuliya-guseva: When we call `client.getCacheManager().getCache`, `cache.get`, `cache.delete`, `cache.save` the SDK opens and closes realm instance, so for creating 1 item we will open and close realm instance additionally 4 times. Opening/closing realm instance aren't quick operations. How will it affect performance?

vinaygahlawat: This will most likely affect performance, but this should be limited to the cases where we are creating entities locally, and not in general. Based on current design, this is necessary for correctness, but we can revisit the performance implications once we take a deeper look at the overall design.

PR #170, opened 2018-05-31
feat: MLIBZ-2470 Test server side delta set use cases

yuliya-guseva:
#### Description
Implement server side delta set tests

#### Changes
Added server side delta set tests
Added fixes: 
1. `pull` returns count of changed items for delta set instead of count of all items from the cache.
2. `find(query)` with delta set returns a correct value, before the `find` returned all items (query was absent)
3. fixed NPE in the list of exceptions in ReadResponse.

#### Tests
Instrumented

### DIFF
--- a/android-lib/src/androidTest/java/com/kinvey/androidTest/store/DeltaSetNewTest.java
+++ b/android-lib/src/androidTest/java/com/kinvey/androidTest/store/DeltaSetNewTest.java
@@ -697,6 +697,38 @@ public class DeltaSetNewTest {
     }
 
 
+    /* error handling*/
+    @Test
+    public void testUnconfiguredCollectionAtTheBackend() throws InterruptedException {
+        store = DataStore.collection(Person.COLLECTION, Person.class, StoreType.SYNC, client);
+        testManager.cleanBackend(store, StoreType.SYNC);
+        store.setDeltaSetCachingEnabled(true);
+        testManager.save(store, new Person(TEST_USERNAME));
+        testManager.push(store);
+        assertEquals(1, testManager.pullCustom(store, client.query()).getResult().getCount());
+        testManager.save(store, new Person(TEST_USERNAME));
+        testManager.push(store);
+        CustomKinveyPullCallback pullCallback = testManager.pullCustom(store, client.query());
+        assertNotNull(pullCallback.getError());
+        assertTrue(pullCallback.getError().getMessage().contains("This feature is not properly configured for this app backend"));
+    }
+
+    /* check that skip and limit is ignored in delta set*/
+    @Test
+    public void testSkipLimit() throws InterruptedException {
+        initDeltaSetCachedCollection(StoreType.SYNC);
+        testManager.cleanBackend(store, StoreType.SYNC);
+        testManager.save(store, new Person(TEST_USERNAME));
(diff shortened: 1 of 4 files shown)


android-lib/src/androidTest/java/com/kinvey/androidTest/store/DeltaSetNewTest.java
```diff
@@ -0,0 +1,766 @@
...
+
+        people = testManager.find(networkStore, emptyQuery).getResult().getResult();
+        person = people.get(0);
+        testManager.delete(networkStore, person.getId());
+        assertEquals(1, testManager.find(networkStore, emptyQuery).getResult().getResult().size());
+    }
+
+    @Test
+    public void testForceNetworkWithQuery() throws InterruptedException {
+        DataStore<Person> networkStore = DataStore.collection(Person.DELTA_SET_COLLECTION, Person.class, StoreType.NETWORK, client);
```
vinaygahlawat: I think the goal of this test is to use a `SYNC` store, but then override the read behavior with `ReadPolicy.FORCE_NETWORK`, to verify that it works as expected, not to just use a `NETWORK` store.

android-lib/src/androidTest/java/com/kinvey/androidTest/store/DeltaSetNewTest.java
```diff
@@ -0,0 +1,766 @@
...
+    }
+
+    @Test
+    public void testFindStoreTypeCache_DeleteOneOfThreeItems() throws InterruptedException {
+        initDeltaSetCachedCollection(StoreType.CACHE);
+        Person person = testManager.save(store, new Person(TEST_USERNAME)).getResult();
+        testManager.save(store, new Person(TEST_USERNAME));
+        testManager.save(store, new Person(TEST_USERNAME_2));
+        assertEquals(2, testManager.find(store, usernameQuery).getResult().getResult().size());
+        testManager.delete(store, person.getId());
```
vinaygahlawat: This test is supposed to have 3 items in the store, but only 2 are asserted. Why is this?

android-lib/src/androidTest/java/com/kinvey/androidTest/store/DeltaSetNewTest.java
yuliya-guseva: Updated test.

android-lib/src/androidTest/java/com/kinvey/androidTest/store/DeltaSetNewTest.java
```diff
@@ -0,0 +1,766 @@
...
+        assertEquals(1, pullCallback.getResult().getCount());
+        updateItem();
+        testManager.push(store);
+        pullCallback = testManager.pullCustom(store, emptyQuery);
+        assertEquals(1, pullCallback.getResult().getCount());
+        testManager.delete(store, emptyQuery);
+        testManager.push(store);
+        pullCallback = testManager.pullCustom(store, emptyQuery);
+        assertEquals(0, pullCallback.getResult().getCount());
+    }
```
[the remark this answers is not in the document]
yuliya-guseva: Added additional checking to the test and to other tests that item saved/removed.

PR #172, opened 2018-06-05
feat: MLIBZ-2533 Make auto-pagination request first if Delta Set and auto-pdagination are enabled

yuliya-guseva:
#### Description
DataStore with AP does not execute Delta Set request

#### Changes
Added auto-pagination request as the first request if delta set and auto-pdagination are enabled

#### Tests
Instrumented

### DIFF
--- a/java-api-core/src/com/kinvey/java/store/BaseDataStore.java
+++ b/java-api-core/src/com/kinvey/java/store/BaseDataStore.java
@@ -59,6 +59,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.FutureTask;
 
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 
 
 public class BaseDataStore<T extends GenericJson> {
@@ -409,7 +410,8 @@ public class BaseDataStore<T extends GenericJson> {
      * should be used with {@link StoreType#SYNC}
      * @param query query to pull the objects
      */
-    public KinveyPullResponse pullBlocking(Query query) throws IOException {
+    @Nonnull
+    public KinveyPullResponse pullBlocking(@Nullable Query query) throws IOException {
         Preconditions.checkArgument(storeType != StoreType.NETWORK, "InvalidDataStoreType");
         Preconditions.checkNotNull(client, "client must not be null.");
         Preconditions.checkArgument(client.isInitialize(), "client must be initialized.");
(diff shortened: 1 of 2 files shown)


java-api-core/src/com/kinvey/java/store/BaseDataStore.java
```diff
@@ -409,7 +410,8 @@ public void pushBlocking() throws IOException {
      * should be used with {@link StoreType#SYNC}
      * @param query query to pull the objects
      */
-    public KinveyPullResponse pullBlocking(Query query) throws IOException {
+    @Nonnull
+    public KinveyPullResponse pullBlocking(@Nullable Query query) throws IOException {
         Preconditions.checkArgument(storeType != StoreType.NETWORK, "InvalidDataStoreType");
```
vinaygahlawat: If we have another `pullBlocking` method that does not take a `Query`, why is the `Query` parameter allowed to be `null`?

java-api-core/src/com/kinvey/java/store/BaseDataStore.java
```diff
@@ -457,104 +460,96 @@ public KinveyPullResponse pullBlocking(Query query, boolean isAutoPagination) th
...
+                executor.shutdown();
+            }
+            response.setListOfExceptions(exceptions);
+            response.setCount(pulledItemCount);
+            if (deltaSetCachingEnabled && lastRequestTime != null) {
+                queryCache.save(new QueryCacheItem(
+                        getCollectionName(),
+                        query.getQueryFilterMap().toString(),
+                        lastRequestTime));
             }
```
vinaygahlawat: When performing auto-pagination, is this guaranteed to be the earliest `lastRequestTime` of the numerous `GET` requests sent?

java-api-core/src/com/kinvey/java/store/BaseDataStore.java
yuliya-guseva: All `pullBclocking` methods have a `Query` in the parameters.

java-api-core/src/com/kinvey/java/store/BaseDataStore.java
yuliya-guseva: It will be the time from the last request which was executed.

vinaygahlawat: Based on the design doc, we should set the `lastRequestTime` to the request time returned from the `_count` request made prior to the auto-paginated GET requests:

https://github.com/Kinvey/java-library/blob/072abd9876591a999f036fd66367069c6912a02c/java-api-core/src/com/kinvey/java/store/BaseDataStore.java#L487

PR #173, opened 2018-06-06
feat: MLIBZ-2534 Send reqular GET after failing Delta Set request

yuliya-guseva:
#### Description
Steps to reproduce:
1. Enable deltaset for a store
2. Do not enable it on the backend
3. Pull 
4. Pull again
Expected result: The first pull will do a regular GET to retrieve all items but the second one will attempt to use _deltaset, since it is not enabled in the backend, the request will fail with Missing Configuration error, and a regular GET will be sent, overriding the current value in _QueryCache with its own X-Kinvey-Request-Start value
Actual result: the first pull sends regular GET, the second attempts to use _deltaset, but it does not attempt to send another GET after receiving the Missing Configuration error

#### Changes
Added error checking if Delta Set request failed, If error is `MissingConfiguration` send reqular GET request.

#### Tests
Instrumented

### DIFF
--- a/java-api-core/src/com/kinvey/java/store/BaseDataStore.java
+++ b/java-api-core/src/com/kinvey/java/store/BaseDataStore.java
@@ -82,6 +82,14 @@ public class BaseDataStore<T extends GenericJson> {
     protected NetworkManager<T> networkManager;
 
     /**
+     * Possible methods where get with delta set is used
+     */
+    private enum METHOD_TYPE {
+        PULL,
+        FIND
+    }
+
+    /**
      * It is a parameter to enable mechanism to optimize the amount of data retrieved from the backend.
      * When you use a Sync or Cache datastore, data requests to the backend
      * only fetch data that changed since the previous update.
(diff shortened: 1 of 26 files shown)


java-api-core/src/com/kinvey/java/store/BaseDataStore.java
```diff
@@ -531,7 +532,18 @@ public KinveyPullResponse pullBlocking(Query query, int pageSize) throws IOExcep
         List<QueryCacheItem> queryCacheItems = queryCache.get(client.query().equals(Constants.QUERY, query.getQueryFilterMap().toString()));
         if (queryCacheItems.size() == 1) { //one is correct number of query cache item count for any request.
             QueryCacheItem cacheItem = queryCacheItems.get(0);
-            KinveyQueryCacheResponse<T> queryCacheResponse = networkManager.queryCacheGetBlocking(query, cacheItem.getLastRequestTime()).execute();
+            KinveyQueryCacheResponse<T> queryCacheResponse;
+            try {
+                queryCacheResponse = networkManager.queryCacheGetBlocking(query, cacheItem.getLastRequestTime()).execute();
+            } catch (KinveyJsonResponseException responseException) {
+                if (responseException.getDetails().getError().equals("MissingConfiguration")
+                        && responseException.getDetails().getDescription().equals("This feature is not properly configured for this app backend. Please configure it through the console first, or contact support for more information.")) {
```
vinaygahlawat: Would be better here to compare against status code instead of error description, since a change in the text of the error description would cause this to fail.

yuliya-guseva: Updated.

PR #179, opened 2018-06-20
feat: MLIBZ-2563 Add updating query cache item if it exists instead of creating new

yuliya-guseva:
#### Description
When a deltaset specific error occurs, deltaset request is not retried on next attempts to pull/sync
Steps to reproduce:

1. Disable deltaset at the backend
2. Pull
3. Pull
4. Pull

Expected result: The first pull makes a regular GET and enter a value in the _QueryCache table, the next one attempts _deltaset request, but it fails for an error, so a regular GET is again made and its value overrides the lastRequestAt value in the _QueryCache table. The third request does exactly the same as the second one

Actual result: The first pull does what is expected, the second uses deltaset and then regular GET, but records a new entry for the query in _QueryCache table, so now it has 2 entries. The third pull directly makes a regular GET instead of first using deltaset. These also make their own entries on the _QueryCache, so after a few pulls, there are a few entries with the same query but with different lastRunAt values.

#### Changes
Added updating the query cache item if it exists instead of creating new.
Added checking that the query cache has only 1 item for the query, else  all items(with the same query) is deleted except the latest one.

#### Tests
Instrumented

### DIFF
--- a/android-lib/src/androidTest/java/com/kinvey/androidTest/store/DeltaSetNewTest.java
+++ b/android-lib/src/androidTest/java/com/kinvey/androidTest/store/DeltaSetNewTest.java
@@ -1274,4 +1274,23 @@ public class DeltaSetNewTest {
         assertEquals(5, pullCallback.getResult().getCount());
         assertEquals(5, testManager.find(store, emptyQuery).getResult().getResult().size());
     }
+
+    @Test
+    public void testDeltaSetDoRequestAfter2FailedAttempts() throws IOException, InterruptedException {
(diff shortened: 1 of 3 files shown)


yuliya-guseva:
In the previous implementation, the QueryCacheItem isn't updated in some cases, so if user has done a lot of `pull` with disabled delta set at the backend, he had an item in QueryCache table for each of this `pull`, because for each `pull` new query cache item was created, therefore I've added logic to clear those duplicates. In the current implementation, QueryCacheTable shouldn't have items with the same query, and this logic won't be used for new items.

PR #180, opened 2018-06-21
feat: MLIBZ-2472 Using 'int' primitive type in class model and store SYNC and CACHE throws an exception

yuliya-guseva:
#### Description
 Using 'int' primitive type in class model and store SYNC and CACHE throws an exception

#### Changes
Added сonverting long to int type when the SDK gets an object from the realm file

#### Tests
Instrumented

### DIFF
--- a/android-lib/src/main/java/com/kinvey/android/cache/ClassHash.java
+++ b/android-lib/src/main/java/com/kinvey/android/cache/ClassHash.java
@@ -817,7 +817,7 @@ public abstract class ClassHash {
                 }
 
                 if (Number.class.isAssignableFrom(info.getType())){
-                    Number n = (Number)dynamic.get(info.getName());
+                    Number n = dynamic.get(info.getName());
                     if (Long.class.isAssignableFrom(info.getType())){
                         ret.put(info.getName(), n.longValue());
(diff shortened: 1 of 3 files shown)


android-lib/src/main/java/com/kinvey/android/cache/ClassHash.java
```diff
@@ -860,8 +860,13 @@ private static int deleteClassData(String collection, DynamicRealm realm, Class<
                             ret.put(info.getName(), c);
                         }
                     }
-                } else {
-                    ret.put(info.getName(), o);
+                } else { //if you are here, then the field is a primitive
+                    // realm keeps int values as a long, so we need to convert it back
+                    if (info.getType().isPrimitive() && info.getType().getSimpleName().equals("int") && o instanceof Long) {
+                        ret.put(info.getName(), ((Long) o).intValue());
```
vinaygahlawat: Converting `long` to `int` could result in losing/truncating data. Is this guaranteed to be a `int` in this case?

yuliya-guseva: Yes, This is be a `int` because we check that this field is `int` before converting`info.getType().getSimpleName().equals("int")`

PR #182, opened 2018-06-26
feat: MLIBZ-2558 Change how sync queue stores request information

yuliya-guseva:
#### Description
Right now, requests for the sync queue are only being saved as a SAVE or DELETE. It would be better to expose out what type of request we are pursuing (POST/PUT/DELETE) in the queue itself, in order to be able to replace temporary IDs with permanent IDs easier.

#### Changes
Changed sync items types from `SAVE or DELETE` to `POST/PUT/DELETE`. `SAVE` type was saved for backward compatibility with previous implementation.

#### Tests
Instrumented

### DIFF
--- a/java-api-core/src/com/kinvey/java/network/NetworkManager.java
+++ b/java-api-core/src/com/kinvey/java/network/NetworkManager.java
@@ -463,7 +463,7 @@ public class NetworkManager<T extends GenericJson> {
     }
 
     public boolean isTempId(T item) {
-        boolean isTempId = true;
+        boolean isTempId = false;
         if (item.get(Constants._ID) != null) {
             try {
                 isTempId = item.get(Constants._ID).toString().startsWith("temp_");
--- a/java-api-core/src/com/kinvey/java/network/NetworkManager.java
+++ b/java-api-core/src/com/kinvey/java/network/NetworkManager.java
@@ -466,7 +466,7 @@ public class NetworkManager<T extends GenericJson> {
         boolean isTempId = true;
         if (item.get(Constants._ID) != null) {
             try {
-                isTempId = item.get(Constants._ID).toString().matches("\\w{8}-\\w{4}-\\w{4}-\\w{4}-\\w{12}");
+                isTempId = item.get(Constants._ID).toString().startsWith("temp_");
             } catch (NullPointerException npe) {
                 // issue with the regex, so do nothing because we default to false
             }
--- a/java-api-core/src/com/kinvey/java/network/NetworkManager.java
+++ b/java-api-core/src/com/kinvey/java/network/NetworkManager.java
@@ -23,6 +23,7 @@ import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
 import com.google.gson.Gson;
 import com.kinvey.java.AbstractClient;
+import com.kinvey.java.Constants;
 import com.kinvey.java.Query;
 import com.kinvey.java.annotations.ReferenceHelper;
 import com.kinvey.java.cache.ICache;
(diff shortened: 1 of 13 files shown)


java-api-core/src/com/kinvey/java/network/NetworkManager.java
```diff
@@ -468,6 +462,18 @@ public String onUnsavedReferenceFound(String collection, GenericJson object) {
         return save;
     }
 
+    public boolean isTempId(T item) {
+        boolean isTempId = true;
+        if (item.get(Constants._ID) != null) {
```
vinaygahlawat: Why is the default value here `true`?  Would it be better if it were `false`, so we only return `true` if we are certain that the format of the `_id` is the temporary format?

java-api-core/src/com/kinvey/java/network/NetworkManager.java
```diff
@@ -468,6 +462,18 @@ public String onUnsavedReferenceFound(String collection, GenericJson object) {
...
+            try {
+                isTempId = item.get(Constants._ID).toString().matches("\\w{8}-\\w{4}-\\w{4}-\\w{4}-\\w{12}");
```
vinaygahlawat: Does this also have to check that the `_id` starts with `temp_`?

java-api-core/src/com/kinvey/java/store/requests/data/PushRequest.java
```diff
@@ -81,7 +86,16 @@ public Void execute() throws IOException {
                     }
                 }
                 try {
-                    syncManager.executeRequest(client, syncRequest);
+                    if (httpVerb == SyncRequest.HttpVerb.POST) {
+                        String tempID = syncRequest.getEntityID().id;
+                        GenericJson result = syncManager.executeRequest(client, syncRequest);
+                        T temp = cache.get(tempID);
+                        temp.set("_id", result.get("_id"));
```
vinaygahlawat: Would be good to use a constant string value here for `_id`, such as `Constants._ID`.

java-api-core/src/com/kinvey/java/sync/SyncManager.java
```diff
@@ -159,35 +161,50 @@ public void enqueueRequest(SyncRequest request) {
         requestCache.save(request);
     }
 
-    public <T extends GenericJson> void enqueueRequest(String collectionName, NetworkManager<T> networkManager, RequestMethod method, String id) throws IOException {
+    public <T extends GenericJson> void enqueueRequest(String collectionName, NetworkManager<T> networkManager, SyncItem.HttpVerb httpMethod, String id) throws IOException {
         ICache<SyncItem> requestCache = cacheManager.getCache(SYNC_ITEM_TABLE_NAME, SyncItem.class, Long.MAX_VALUE);
```
vinaygahlawat: Are some of these API changes potentially breaking changes to the public API?

java-api-core/src/com/kinvey/java/network/NetworkManager.java
yuliya-guseva: Yes, this has to check that the `_id` starts with ` temp_`, but it was the feature from another branch which was merged in the current branch today, so currently the SDK already checks ` temp_`. If a user had items which weren't synced the items will be pushed using `PUT`, but all new items will be pushed with `POST` request. The PR has instrumented test with such case.

java-api-core/src/com/kinvey/java/store/requests/data/PushRequest.java
yuliya-guseva: I agree. It was updated.

java-api-core/src/com/kinvey/java/network/NetworkManager.java
yuliya-guseva: You are right. And if the SDK gets `true` from this method then the POST request will be made and if it was not temp id then we will have duplicate items at the backend with different ids. Updated.

java-api-core/src/com/kinvey/java/sync/SyncManager.java
yuliya-guseva: We don't have any notes about this method or this class at the devcenter. It's class for internal SDK using. I'm not sure that we have to add this changes to `breaking changes` for release notes. These changes don't affect classes and methods which exist in devcenter notes.

PR #184, opened 2018-06-28
feat: MLIBZ-2577 Using Auto-pagination does not delete all items in the cache

yuliya-guseva:
#### Description
Steps to reproduce:

1. Create 5 items with some property value - e.g. name = name1
2. Create 3000 items with another value - e.g. name = name2
3. Pull() or Sync() with AP = 1000
4. Check the db - all 3005 items should be there
5. Delete all items with name=name2 from the backend
6. Pull() or Sync() with AP=1000

Expected result: All 3000 items with name=name2 are deleted from the cache

Actual result: The AP pull gets only the 5 items with name = name1, which were not deleted from the backend. Not all items with name =name2, however, are deleted from cache. The number of actually deleted items from the cache = the AP size(1000) - the result of the _count request from the AP pull() - in our case 5. So from all 3005 items, after the pull from step 6, there are 2005 items in the cache, the next pull with AP = 1000 will delete another 1000-5 items.

#### Changes
Added deleting all items by query from the cache before auto-paginated concurrent pull. 

#### Tests
Instrumented

### DIFF
--- a/java-api-core/src/com/kinvey/java/store/BaseDataStore.java
+++ b/java-api-core/src/com/kinvey/java/store/BaseDataStore.java
@@ -502,12 +502,14 @@ public class BaseDataStore<T extends GenericJson> {
         List<FutureTask<PullTaskResponse>> tasks;
         NetworkManager.Get pullRequest;
         FutureTask<PullTaskResponse> ft;
+        cache.delete(query.setSkip(0).setLimit(0));// To be sure that skip and limit are 0,
+        // because in next lines custom skip and limit are set anyway
         for (int i = 0; i < totalPagesNumber; i += batchSize) {
             executor = Executors.newFixedThreadPool(batchSize);
             tasks = new ArrayList<>();
             do {
                 query.setSkip(skipCount).setLimit(pageSize);
-                pullRequest = networkManager.pullBlocking(query);
+                pullRequest = networkManager.getBlocking(query);
                 skipCount += pageSize;
                 try {
                     ft = new FutureTask<PullTaskResponse>(new CallableAsyncPullRequestHelper(pullRequest, query));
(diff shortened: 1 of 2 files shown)


java-api-core/src/com/kinvey/java/store/BaseDataStore.java
```diff
@@ -502,12 +502,14 @@ private KinveyPullResponse pullBlockingPaged(@Nonnull Query query, int pageSize)
         List<FutureTask<PullTaskResponse>> tasks;
         NetworkManager.Get pullRequest;
         FutureTask<PullTaskResponse> ft;
+        cache.delete(query.setSkip(0).setLimit(0));// To be sure that skip and limit are 0,
```
vinaygahlawat: Is there a possibility here that `skip` and `limit` are not 0?

java-api-core/src/com/kinvey/java/store/BaseDataStore.java
```diff
@@ -536,6 +537,7 @@ private KinveyPullResponse pullBlockingPaged(@Nonnull Query query, int pageSize)
             }
             executor.shutdown();
         }
+        query.setSkip(0).setLimit(0); // To set back default value of skip and limit
```
vinaygahlawat: What is the reason for setting the default values on the `skip` and `limit` modifiers here?

yuliya-guseva: How I wrote above, the SDK sets own `skip` and `limit` for auto-pagination, and after pagination pull user's query will have `skip` and `limit` from the last page of pagination request even if the user didn't set `skip` or `limit` before calling `pull` with auto-pagination. We can exclude this from the PR, but I think it's more user-friendly.

vinaygahlawat: That's fine. Another potential way of doing this in the future would be to clone the user's query, and apply `skip` and `limit` modifiers on the clone for AP purposes. That way, the user's query remains untouched.

PR #187, opened 2018-07-11
feat: MLIBZ-2602 Update README.md

yuliya-guseva:
#### Description
Outline for how to build the SDK, as well as build a solution to use TLS v1.2 in the SDK.

#### Changes
Actualized build steps.
Added how to enable support for TLSv1.1-1.2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,16 +1,14 @@
 Kinvey Java Library
 ======
 
-This library is a standalone library designed for all java environments.
-The library acts as a client for the Kinvey REST api and can be used for
-building Android apps and Java6 server applications.
+This library is a standalone library designed for all java environments. The library acts as a client for the Kinvey REST API and can be used for building Android apps and Java6 server applications.
 
 It is recommended to use either IntelliJ or Android Studio. Eclipse is NOT recommended.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,115 +1,157 @@
...
-rm -r <devcenter.home>/content/reference/android/api/*
-cd <project.home> 
-mvn -Pdev javadoc:javadoc install
+./gradlew clean release
 ```
 
-### Release
+The command should generate a .zip archive in `/release/zipped` containing .aar and .jar files.
 
+## How to enable TLSv1.1/TLSv1.2 on Android v4.1-v4.4.4
```
boevski: @vinaygahlawat Should we follow suit and downplay v1.1 as in the other documents? (This should also require tweaking the example code a bit to not enable this version.)

vinaygahlawat: @boevski I think in the case of the example that it is still ok to discuss 1.1 and 1.2, as this is just enabling these protocol versions, which is correct. As you mentioned before, we can still emphasize 1.2 in our instructions, as long as we still allow for both 1.1 and 1.2.

PR #195, opened 2018-08-30
feat: MLIBZ-2653 UserStore#get doesn't return User object fix

yuliya-guseva:
#### Description
The goal of the task is to fix the code to return an object of User for method 'UserStore#get'

#### Changes
'UserStore#get' method returned null in any cases, it was fixed for return User object.
 
#### Tests
Instrumented

### DIFF
--- a/android-lib/src/androidTest/java/com/kinvey/androidTest/store/UserStoreTest.java
+++ b/android-lib/src/androidTest/java/com/kinvey/androidTest/store/UserStoreTest.java
@@ -375,6 +375,30 @@ public class UserStoreTest {
     }
 
     @Test
+    public void get() throws InterruptedException {
+        login(USERNAME, PASSWORD);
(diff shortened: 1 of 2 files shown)


android-lib/src/androidTest/java/com/kinvey/androidTest/store/UserStoreTest.java
```diff
@@ -374,6 +374,30 @@ public void testLogin() throws InterruptedException {
         assertNull(logout(client).error);
     }
 
+    @Test
+    public void get() throws InterruptedException {
+        login(USERNAME, PASSWORD);
+        DefaultKinveyClientCallback callback = get(client.getActiveUser().getId());
+        assertNull(callback.error);
+        assertNotNull(callback.result);
+    }
```
vinaygahlawat: In this test, we should also verify that we get back the expected `User` object, in addition to checking for a non-null result.

PR #201, opened 2018-10-11
MLIBZ-2601 Give feedback  Add tests for Query class methods which aren't supported by the cache but have to work with StoreType.NETWORK

yuliya-guseva:
#### Description
The goal of the task is to add tests for Query class methods which aren't supported by the cache but have to work with StoreType.NETWORK

#### Changes
Added test class QueryNetworkTest.java and model class Location.java

#### Tests
Instrumented

### DIFF
--- a/android-lib/src/androidTest/java/com/kinvey/androidTest/store/data/network/QueryNetworkTest.java
+++ b/android-lib/src/androidTest/java/com/kinvey/androidTest/store/data/network/QueryNetworkTest.java
@@ -1,0 +1,444 @@
+package com.kinvey.androidTest.store.data.network;
+
+import android.content.Context;
+import android.os.Message;
+import android.support.test.InstrumentationRegistry;
+import android.support.test.runner.AndroidJUnit4;
+import android.test.RenamingDelegatingContext;
+import android.test.suitebuilder.annotation.SmallTest;
+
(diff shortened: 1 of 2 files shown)


android-lib/src/androidTest/java/com/kinvey/androidTest/store/data/network/QueryNetworkTest.java
```diff
@@ -0,0 +1,444 @@
...
+        DataStore<Person> store = DataStore.collection(Person.COLLECTION, Person.class, StoreType.NETWORK, client);
+        clearBackend(store);
+        client.getSyncManager().clear(Person.COLLECTION);
+        Person person = createPerson(TEST_USERNAME);
+        DefaultKinveyClientCallback saveCallback = save(store, person);
+        assertNotNull(saveCallback.result);
+        assertNull(saveCallback.error);
+        assertNotNull(saveCallback.result.getId());
+        Query query = client.query();
+        query = query.startsWith(USERNAME, "^T");
```
vinaygahlawat: What is the `^` for in `"^T"`?

yuliya-guseva: It is the part of regular expression

PR #202, opened 2018-10-12
MLIBZ-2643 Time from Date object in milliseconds before saving to the DataStore isn't equaled time after getting the entity from the DataStore

yuliya-guseva:
#### Description
The goal of the task is to find why time from Date object in milliseconds before saving to the DataStore isn't equaled time after getting the entity from the DataStore

#### Changes
Add test for entity with date and fix RealmCache

#### Tests
Instrumented

### DIFF
--- a/android-lib/src/main/java/com/kinvey/android/cache/RealmCache.java
+++ b/android-lib/src/main/java/com/kinvey/android/cache/RealmCache.java
@@ -508,6 +508,7 @@ public class RealmCache<T extends GenericJson> implements ICache<T> {
 
         item.set(ID, clone.get(ID));
 
+        item.remove(ClassHash.TTL);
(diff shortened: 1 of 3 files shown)


android-lib/src/main/java/com/kinvey/android/cache/RealmCache.java
```diff
@@ -501,13 +501,14 @@ void checkAclKmdFields(DynamicRealm realm) {
 
     private String insertOrUpdate(T item, DynamicRealm mRealm){
 
-        T clone = (T)item.clone();
+        T clone = (T)item;
```
vinaygahlawat: What is the affect of not cloning the `item`? And why do we have a `clone` variable, if it is just the `item`?

PR #203, opened 2018-11-06
MLIBZ-2647 Self-reference data model with an array

yuliya-guseva:
#### Description
The goal of the task is to find why the SDK throws the exception after getting saved entity, which has an array with self-reference type in entity data model.

#### Changes
Fix the condition in ClassHash.createSchemeFromClass

#### Tests
Instrumented

### DIFF
--- a/android-lib/src/main/java/com/kinvey/android/cache/ClassHash.java
+++ b/android-lib/src/main/java/com/kinvey/android/cache/ClassHash.java
@@ -380,7 +380,7 @@ public abstract class ClassHash {
             if (fieldInfo == null){
                 continue;
             }
-            if (fieldInfo.getType().isArray() || Collection.class.isAssignableFrom(fieldInfo.getType())){
+            if (Collection.class.isAssignableFrom(fieldInfo.getType())){
                 Class underlying = getUnderlying(f);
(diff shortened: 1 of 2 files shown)


android-lib/src/main/java/com/kinvey/android/cache/ClassHash.java
```diff
@@ -380,7 +380,7 @@ private static RealmObjectSchema createSchemeFromClass(String name, DynamicRealm
             if (fieldInfo == null){
                 continue;
             }
-            if (fieldInfo.getType().isArray() || Collection.class.isAssignableFrom(fieldInfo.getType())){
```
vinaygahlawat: @yuliya-guseva Can you explain more why this condition was causing an exception?

yuliya-guseva: This class works with Realm tables and contains a complicated logic.
This part of the condition was logically incorrect because we go into this part of the code only if the field is an array and the type of the array inherits `GenericJson`. In this case, the `RealmObjectSchema.addRealmListField` method is called and the field is created in the table in the form of ArrayList. But for the array this method should not be called.

PR #205, opened 2018-11-15
MLIBZ 2723 increase the code coverage of all files where values are 0 - 10% in the com.kinvey.android package

yuliya-guseva:
#### Description
The goal of the task increase the code coverage of all files where values are 0 - 10% in the com.kinvey.android package 

#### Changes
Added some tests for files where values are 0 - 10%

#### Tests
Instrumented


android-lib/src/androidTest/java/com/kinvey/androidTest/store/data/network/AsyncLinkedNetworkManagerTest.java
```diff
@@ -73,6 +78,48 @@ public void testConstructor() {
...
+        androidMimeTypeFinder.getMimeType(fileMetaData, (File) null);
+        assertEquals(fileMetaData.getMimetype(), "application/octet-stream");
+    }
+
+    @Test
+    public void testMimeTypeFinderNullMimetype() {
+        AndroidMimeTypeFinder androidMimeTypeFinder = new AndroidMimeTypeFinder();
+        FileMetaData fileMetaData = new FileMetaData();
+        androidMimeTypeFinder.getMimeType(fileMetaData);
+        assertEquals(fileMetaData.getMimetype(), "application/octet-stream");
```
vinaygahlawat: How is this testing for a null Mimetype?

yuliya-guseva: The method has the following code:
`@Override`
   ` public void getMimeType(FileMetaData metaData) {`
     `   if (metaData.getMimetype() != null && metaData.getMimetype().length() > 0) {`
         `   return;`
       ` }`
     `   String mimetype = null;`

        if (metaData.getFileName() != null) {
            int dotIndex = metaData.getFileName().lastIndexOf("");

            if (dotIndex > 0 && dotIndex + 1 < metaData.getFileName().length()) {
                mimetype = MimeTypeMap.getSingleton().getMimeTypeFromExtension(metaData.getFileName().substring(dotIndex + 1, metaData.getFileName().length()));
            }
        }

        if (mimetype == null) {
            mimetype = "application/octet-stream";
        }
        metaData.setMimetype(mimetype);
    }`
If we put `metaData` with `metaData.getMimetype() == null` and `metaData.getFileName() == null` in the method, we will have `fileMetaData.getMimetype().equals( "application/octet-stream")`

PR #208, opened 2018-12-10
MLIBZ-2762 Deprecate `loginWithAuthorizationCodeAPI()` method for automated authorization grant flow

yuliya-guseva:
#### Description
The goal of the task is deprecate loginWithAuthorizationCodeAPI() method for automated authorization grant flow

#### Changes
Deprecated loginWithAuthorizationCodeAPI()  in favor of a `loginWithMIC()` method 

#### Tests
Instrumented


b-stolyarov:
I think we do not need to make unit test changes and create new tests, because we have tests for old methods and they cover new method because now we call `LoginWithMIC` in old methods

PR #209, opened 2018-12-20
MLIBZ-2080 not() operator - should it transform Mongo operators or add $not Mongo operator

yuliya-guseva:
#### Description
The goal of the task is to remove operator replacement logic

#### Changes
 Removed operator replacement logic

#### Tests

Instrumented

### DIFF
--- a/java-api-core/src/com/kinvey/java/query/MongoQueryFilter.java
+++ b/java-api-core/src/com/kinvey/java/query/MongoQueryFilter.java
@@ -180,8 +180,9 @@ public class MongoQueryFilter implements QueryFilter, Serializable {
         private Boolean negate(LinkedHashMap<String, Object> map) {
 
             boolean processParent = false;
-            LinkedHashMap<String,Object> newMap = new LinkedHashMap<String, Object>();
+            LinkedHashMap<String, LinkedHashMap<String, Object>> newMap = new LinkedHashMap<>();
             LinkedHashMap<String,Object> tempMap = (LinkedHashMap<String,Object>) map.clone();
+            LinkedHashMap<String, Object> element = new LinkedHashMap<>();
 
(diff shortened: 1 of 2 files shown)


java-api-core/src/com/kinvey/java/query/MongoQueryFilter.java
```diff
@@ -202,22 +203,28 @@ private Boolean negate(LinkedHashMap<String, Object> map) {
...
                             break;
                         case IN :
-                            newMap.put("$nin", obj);
+                            element.put("$in", obj);
+                            newMap.put("$not", element);
                             break;
                         case NOTIN :
-                            newMap.put("$in", obj);
+                            element.put("$nin", obj);
+                            newMap.put("$not", element);
```
vinaygahlawat: Would this effectively make this filter an `$in` filter? If so, this is not the desired effect of the `NOTIN` operator. I believe that this should still put an `$in` in the `element`, and then apply the `$not` as with the other operations. Either that, or it should just be the `$nin` operation that gets placed in `newMap`, but there may be differences in the results of both of these, and we should understand what they are.

b-stolyarov: This changes is located in negate() method. So expected result of  negate `case:NOTIN`  - `IN`. 
Previously, we just inserted into the map `$in` Now the structure of map has changed and we need to write down the negation explicitly. We already have tests for method `$in`. We changed implementation of operation `$not` but result of this operation have to stay the same. So if old tests for operator `$not` are passed, we can think that the operator is working right

vinaygahlawat:
@b-stolyarov LGTM, thanks for the explanation on the `negate` method! It looks like tests are not passing, please correct this prior to merging this.

PR #212, opened 2019-01-17
Access Token path might be affected by the MIC API version.

bilger-progress:
#### Description
Currently the path for retrieving the access token is hardcoded and changes to the MIC API version do not get applied. This fix makes sure that the MIC API version that's set from the code is being applied.

#### Changes
`mKinveyClient.setMICApiVersion("3");` this call now makes sure that the token request goes to the correct path.

#### Tests
The test reflecting this from `BaseUserTest.java` file has been fixed and adjusted.

### DIFF
--- a/java-api-core/src/com/kinvey/java/store/requests/user/GetMICAccessToken.java
+++ b/java-api-core/src/com/kinvey/java/store/requests/user/GetMICAccessToken.java
@@ -20,17 +20,20 @@ import com.google.api.client.http.HttpContent;
 import com.google.api.client.json.GenericJson;
 import com.kinvey.java.core.AbstractKinveyClientRequest;
 import com.kinvey.java.store.UserStoreRequestManager;
+import com.kinvey.java.AbstractClient;
 
 /**
  * Created by Prots on 2/12/16.
  */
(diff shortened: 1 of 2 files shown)


b-stolyarov:
Previously we did not have `MICApiVersion` in the path of request, Why should we add it now? Also target branch for PR need be `indev`

vinaygahlawat:
@b-stolyarov In general, we need to have a way of setting the MIC API Version in the URL, because not having it will mean that we always use the older version of MIC.

bilger-progress:
@b-stolyarov, At this point we are able to set different API versions that the SDK can use. Unfortunately, probably is got missed, and the URL for the MIC API was hardcoded, which currently leads to requests being made to the older API version, which might get deprecated in the future. On the other hand, I now made sure that the target branch is `indev`.

bilger-progress:
@b-stolyarov @vinaygahlawat Guys, since Bogdan already re-created the PR and merged it to `indev` can I go ahead and close this PR and delete the branch?

bilger-progress:
I'll go ahead and delete the branch as well. Thanks.

PR #218, opened 2019-02-13
feat: MLIBZ-2878 SPIKE: Research client-side delays in request

b-stolyarov:
#### Description
Research client-side delays in request transmission

#### Changes
Added logs from the time when `save` is called to the time when we print the REQUEST output in the logger

#### Tests
Instrumented

### DIFF
--- a/java-api-core/src/com/kinvey/java/network/NetworkManager.java
+++ b/java-api-core/src/com/kinvey/java/network/NetworkManager.java
@@ -465,10 +465,13 @@ public class NetworkManager<T extends GenericJson> {
 
         boolean bRealmGeneratedId = isTempId(entity);
 
+        Logger.INFO("Start choosing PUT or POST request");
         if (sourceID != null && !bRealmGeneratedId) {
+            Logger.INFO("Start for preparing new Save(entity, myClass, sourceID, SaveMode.PUT)");
             save = new Save(entity, myClass, sourceID, SaveMode.PUT);
             Logger.INFO("Finish for preparing new Save(entity, myClass, sourceID, SaveMode.PUT)");
         } else {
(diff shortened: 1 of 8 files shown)


java-api-core/src/com/kinvey/java/network/NetworkManager.java
```diff
@@ -458,16 +467,20 @@ public String onUnsavedReferenceFound(String collection, GenericJson object) {
 
         if (sourceID != null && !bRealmGeneratedId) {
             save = new Save(entity, myClass, sourceID, SaveMode.PUT);
+            Logger.INFO("Finish for preparing new Save(entity, myClass, sourceID, SaveMode.PUT)");
         } else {
             save = new Save(entity, myClass, SaveMode.POST);
+            Logger.INFO("Finish for preparing new Save(entity, myClass, SaveMode.POST)");
         }
 
         client.initializeRequest(save);
```
vinaygahlawat: Let's add a log for just before the request initialization begins.

java-api-core/src/com/kinvey/java/network/NetworkManager.java
```diff
@@ -458,16 +467,20 @@ public String onUnsavedReferenceFound(String collection, GenericJson object) {
 
         if (sourceID != null && !bRealmGeneratedId) {
```
vinaygahlawat: Let's add a log statement before this `if` block, so we know when the `ReferenceHelper.processReferences` operation finishes.

java-api-core/src/com/kinvey/java/network/NetworkManager.java
b-stolyarov: Added

PR #220, opened 2019-03-05
MLIBZ-2891 Use multi-threading model to handle network requests

b-stolyarov:
#### Description
Use multi-threading model to handle network requests

#### Changes
Added logic with pool of threads for requests

#### Tests
Instrumented


b-stolyarov:
Performance gains depends on the number of threads that we set in `kinvey.properties` file for pool of threads. For example after my test for saving 1500 entities I have about 5 minutes delay with 1 thread and I have delay about 25 seconds for pool with 15 threads

PR #221, opened 2019-03-12
MLIBZ-2680 Migrate from GCM to FCM

b-stolyarov:
#### Description
MIgration from GCM to FCM

#### Changes
Migrated project to FCM

#### Tests
Instrumented


android-lib/build.gradle
```diff
@@ -150,10 +150,11 @@ dependencies {
   compile 'com.android.support:appcompat-v7:28.0.0'
   compile 'com.android.support:support-v4:28.0.0'
   compile 'org.codehaus.jackson:jackson-core-asl:1.9.12'
-  compile 'com.google.android.gms:play-services:10.2.1'
   compile 'com.google.android.gms:play-services-ads:10.2.1'
   compile 'com.google.android.gms:play-services-identity:10.2.1'
-  compile 'com.google.android.gms:play-services-gcm:10.2.1'
+  compile 'com.google.android.gms:play-services-gcm:16.0.0'
```
heyzooi: @b-stolyarov I don't think we to keep GCM

android-lib/src/main/java/com/kinvey/android/push/KinveyFCMService.java
```diff
@@ -0,0 +1,75 @@
...
+            Log.d(TAG, "Message data payload: " + remoteMessage.getData());
+            processMessage((String) remoteMessage.getData().values().toArray()[0]);
+        }
+        if (remoteMessage.getData() == null && remoteMessage.getNotification() != null) {
+            Log.d(TAG, "Message Notification Body: " + remoteMessage.getNotification().getBody());
+            processMessage(remoteMessage.getNotification().getBody());
+        }
+    }
+
+    public abstract void processMessage(String r);
```
heyzooi: I believe the previous interface is called onMessage(), so let's try to keep the same interface

android-lib/src/main/java/com/kinvey/android/push/KinveyFCMService.java
```diff
@@ -0,0 +1,75 @@
...
+     * Persist token to third-party servers.
+     *
+     * Modify this method to associate the user's FCM InstanceID token with any server-side account
+     * maintained by your application.
+     *
+     * @param token The new token.
+     */
+    @Override
+    public void onNewToken(String token) {
+        Log.d(TAG, "New FCM InstanceID token");
```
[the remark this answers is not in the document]
heyzooi: @b-stolyarov I don't think we should expect to developers implement the `onNewToken()` method since it will cause problems if the developer forget to do that. Better safe then sorry in this case I believe and go ahead and do this ourselves.

PR #225, opened 2019-04-12
MLIBZ-2933 Create overload of `loginWithMIC` without `temporaryURI` parameter

b-stolyarov:
#### Description
Create overload of `loginWithMIC` without `temporaryURI` parameter

#### Changes
Added an overload that removes this parameter, and deprecated a version of `loginWithMIC` that takes a `temporaryURI` 

#### Tests
Instrumented

### DIFF
--- a/android-lib/src/main/java/com/kinvey/android/store/UserStore.java
+++ b/android-lib/src/main/java/com/kinvey/android/store/UserStore.java
@@ -156,7 +156,7 @@ public class UserStore {
      * @param client {@link Client} an instance of the client
      * @param callback {@link com.kinvey.java.core.KinveyClientCallback<User>} the callback
      * @throws IOException
-     * @deprecated Use {@link #loginWithMIC(Client, String, String, KinveyMICCallback)} or {@link #loginWithMIC(AbstractClient, String, String, String, String, KinveyUserCallback)}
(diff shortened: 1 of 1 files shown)


android-lib/src/main/java/com/kinvey/android/store/UserStore.java
```diff
@@ -807,14 +807,28 @@ public static void loginWithAuthorizationCodeAPI(@NonNull AbstractClient client,
...
      */
     public static void loginWithAuthorizationCodeAPI(@NonNull AbstractClient client, @NonNull String username,
                                                      @NonNull String password, @Nullable String clientId,
                                                      @NonNull String redirectURI, @NonNull KinveyUserCallback<User> callback){
         loginWithMIC(client, username, password, clientId,redirectURI, callback);
     }
 
+    /***
+     *
+     * Login with the MIC service, using the oauth flow.  This method provides direct login, without rending a login page and without redirectURI.
```
vinaygahlawat: This first sentence should read "Login with the MIC service, using the resource owner grant flow."

b-stolyarov: Ok, fixed

PR #226, opened 2019-04-16
MLIBZ-2823 Correctly handle missing or null refresh token

b-stolyarov:
#### Description
Correctly handle missing or null refresh token

#### Changes
added checking for `refresh_token==null`

#### Tests
Instrumented

### DIFF
--- a/android-lib/src/main/java/com/kinvey/android/store/UserStore.java
+++ b/android-lib/src/main/java/com/kinvey/android/store/UserStore.java
@@ -1116,7 +1116,9 @@ public class UserStore {
             UserStoreRequestManager requestManager = new UserStoreRequestManager(client, createBuilder(client));
             requestManager.setMICRedirectURI(redirectURI);
             GenericJson result = requestManager.getMICToken(token, clientId).execute();
-
+            if (result.get(ACCESS_TOKEN) == null || result.get(REFRESH_TOKEN) == null) {
+                throw new KinveyException("Invalid Access or Refresh MIC token");
+            }
             T ret =  BaseUserStore.loginMobileIdentity(result.get("access_token").toString(), client);
 
             Credential currentCred = client.getStore().load(client.getActiveUser().getId());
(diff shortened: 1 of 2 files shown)


android-lib/src/main/java/com/kinvey/android/store/UserStore.java
```diff
@@ -1116,7 +1116,9 @@ protected T executeAsync() throws IOException {
             UserStoreRequestManager requestManager = new UserStoreRequestManager(client, createBuilder(client));
             requestManager.setMICRedirectURI(redirectURI);
             GenericJson result = requestManager.getMICToken(token, clientId).execute();
-
+            if (result.get(ACCESS_TOKEN) == null || result.get(REFRESH_TOKEN) == null) {
```
vinaygahlawat: It is allowed that you can have an `access_token` but not necessarily a `refresh_token`, so we cannot have this `||` logic here.

android-lib/src/main/java/com/kinvey/android/store/UserStore.java
```diff
@@ -1150,6 +1152,9 @@ protected T executeAsync() throws IOException {
             UserStoreRequestManager requestManager = new UserStoreRequestManager(client, createBuilder(client));
             requestManager.setMICRedirectURI(redirectURI);
             GenericJson result = requestManager.getOAuthToken(clientId, username, password).execute();
+            if (result.get(ACCESS_TOKEN) == null || result.get(REFRESH_TOKEN) == null) {
```
vinaygahlawat: Same comment as above.

android-lib/src/main/java/com/kinvey/android/store/UserStore.java
b-stolyarov: Fixed

android-lib/src/main/java/com/kinvey/android/store/UserStore.java
b-stolyarov: Fixed

PR #229, opened 2019-05-17
MLIBZ - 2956 - Add new DataStore.count method with query object

b-stolyarov:
#### Description
Add new DataStore.count method with query object

#### Changes
Added new DataStore.count method with query object

#### Tests
Instrumented

### DIFF
--- a/java-api-core/src/com/kinvey/java/network/NetworkManager.java
+++ b/java-api-core/src/com/kinvey/java/network/NetworkManager.java
@@ -424,6 +424,16 @@ public class NetworkManager<T extends GenericJson> {
     }
 
     /**
+     * Get count of queried items in the collection
+     * @return GetCount request
+     * @throws IOException
+     */
+    public GetCount getCountBlocking(Query query) throws IOException {
+        GetCount getCount = new GetCount(query);
+        client.initializeRequest(getCount);
+        return getCount;
+    }
+    /**
      * Save (create or update) an entity to a collection.
      *
      * @param entity Entity to Save
@@ -652,7 +662,7 @@ public class NetworkManager<T extends GenericJson> {
         protected String limit;
         @Key("skip")
         protected String skip;
-        
+
 
         @Key
         private String fields = "_id,_kmd";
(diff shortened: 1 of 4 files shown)


java-api-core/src/com/kinvey/java/network/NetworkManager.java
```diff
@@ -950,10 +960,18 @@ public T execute() throws IOException {
      */
     public class GetCount extends AbstractKinveyReadHeaderRequest<KinveyCountResponse> {
 
-        private static final String REST_PATH = "appdata/{appKey}/{collectionName}/_count";
```
vinaygahlawat: Are we removing this `REST_PATH`? Do we still need it for the `GetCount()` where no query is passed?

java-api-core/src/com/kinvey/java/store/BaseDataStore.java
```diff
@@ -315,6 +315,23 @@ public Integer count(KinveyCachedClientCallback<Integer> cachedCallback) throws
         return new ReadCountRequest<T>(cache, networkManager, this.storeType.readPolicy, null, client.getSyncManager()).execute().getCount();
     }
 
+    /**
+     * Get count of queried items in collection
+     * @param cachedCallback is using with StoreType.CACHE to get items count in collection
+     * @return count of queried items in collection
+     */
+    @Nonnull
+    public Integer count(KinveyCachedClientCallback<Integer> cachedCallback, Query query) throws IOException {
```
vinaygahlawat: Since we are deprecating the `.Cache` store type, I do not think that we need to have a method that takes a `cachedCallback` as a parameter.

java-api-core/src/com/kinvey/java/network/NetworkManager.java
b-stolyarov: New `REST_PATH` works either with query and without it.

java-api-core/src/com/kinvey/java/store/BaseDataStore.java
b-stolyarov: We have this logic for count request without query, and I think we need it for new case, because I guess we can delete `KinveyCachedClientCallback` logic only when we will delete `CACHE` store type from project, WDYT?

vinaygahlawat: This is a good point. But in this case, since this is new functionality, I believe that we can introduce this method without having the cache callback. If users want to use this method, they should use it with an actively-supported data store.

PR #231, opened 2019-05-30
MLIBZ - 3014 - Implement multi-record insert

b-stolyarov:
#### Description
Implement multi-record insert

#### Changes
Modified save method in DataSore. Added feature to save multiple items in one request if sdk API version is 5.

#### Tests
Instrumented


android-lib/src/main/java/com/kinvey/android/store/DataStore.java
```diff
@@ -542,13 +546,29 @@ public void save(@NonNull T entity, @NonNull KinveyClientCallback<T> callback)
...
         Preconditions.checkNotNull(entities, "Entity cannot be null.");
         Logger.INFO("Calling DataStore#save(listObjects)");
         new SaveListRequest(entities, callback).execute();
     }
 
+    private void saveBatch(@NonNull List<T> entities, @NonNull KinveyClientCallback<List<T>> callback)  {
+        Preconditions.checkNotNull(client, "client must not be null");
+        Preconditions.checkArgument(client.isInitialize(), "client must be initialized.");
+        Preconditions.checkNotNull(entities, "Entity cannot be null.");
+        Logger.INFO("Calling DataStore#saveBatch(listObjects)");
```
vinaygahlawat: Need another Precondition here if the number of items in the list of entities is greater than 100.

java-api-core/src/com/kinvey/java/model/KinveyError.java
```diff
@@ -0,0 +1,37 @@
+package com.kinvey.java.model;
+
+import com.google.api.client.util.Key;
+
+public class KinveyError {
```
vinaygahlawat: The name of this class should be more specific, since it is for multi-insert errors.

android-lib/src/main/java/com/kinvey/android/store/DataStore.java
b-stolyarov: I've added new precondition `checkPositionIndex` that will stop execution if user will try to send more than 100 items per request.

java-api-core/src/com/kinvey/java/model/KinveyError.java
b-stolyarov: Renamed KinveyError class to KinveyBatchInsertError.

PR #238, opened 2019-06-29
MLIBZ-3057: Multi insert for push logic

b-stolyarov:
#### Description
Multi-insert support in `push` classes logic.

#### Changes
Added multi-insert support in `push` classes logic.

#### Tests
Instrumented.


android-lib/src/main/java/com/kinvey/android/async/AsyncBatchPushRequest.java
```diff
@@ -0,0 +1,267 @@
...
+                    callback.onFailure(e);
+                }
+                callback.onProgress(pushResponse.getSuccessCount(), fullCount);
+            }
+        }
+        return resultSingleItems;
+    }
+
+    private KinveySyncSaveBatchResponse processBatchSyncRequest(List<SyncItem> syncItems, List<T> saveItems) throws IOException {
+        SyncRequest syncRequest = manager.createSaveBatchSyncRequest(collection, networkManager, saveItems);
```
heyzooi: I believe you need to split this request into multiple requests if the length/size of saveItems is greater than `100`

android-lib/src/main/java/com/kinvey/android/async/AsyncBatchPushRequest.java
```diff
@@ -0,0 +1,267 @@
+/*
```
heyzooi: since this is a new class, do you think we should implement in Kotlin instead? And Also `KinveyPushBatchResponse`, `KinveySyncSaveBatchResponse`, `KinveyUpdateObjectsResponse`, `KinveyUpdateSingleItemError`

android-lib/src/main/java/com/kinvey/android/store/DataStore.java
```diff
@@ -682,12 +683,24 @@ public void delete(@NonNull Query query, @NonNull KinveyDeleteCallback callback)
      *
      * @param callback KinveyPushCallback
      */
-    public void push(@NonNull KinveyPushCallback callback){
+    public void push(@NonNull KinveyPushCallback callback) {
         Preconditions.checkNotNull(client, "client must not be null");
         Preconditions.checkArgument(client.isInitialize(), "client must be initialized.");
+        if (KINVEY_API_VERSION_5.equals(kinveyApiVersion)) {
```
heyzooi: better check if `kinveyApiVersion` is `greater or equal than` instead of `equal`

java-api-core/src/com/kinvey/java/model/KinveyBatchInsertError.java
```diff
@@ -27,11 +27,11 @@ public void setCode(long code) {
         this.code = code;
     }
 
-    public long getIndex() {
+    public int getIndex() {
```
heyzooi: since this is a `public` member, this will be a breaking change right? Is there any way to avoid this breaking change?

java-api-core/src/com/kinvey/java/model/KinveyUpdateSingleItemError.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @Key
+    private GenericJson entity;
+    @Key
+    private long code;
+    @Key
+    private String errmsg;
+
+    public String getErrmsg() {
```
heyzooi: since this is already an Exception/Error class, I think it's fine to give this a better name like `message` instead of `errmsg`. What do you think?

java-api-core/src/com/kinvey/java/sync/dto/SyncRequest.java
```diff
@@ -49,9 +49,11 @@
         }
 
         public static HttpVerb fromString(String verb){
-            for (HttpVerb v : HttpVerb.values()){
-                if (v.query.equals(verb)){
-                    return v;
+            if (verb != null) {
+                for (HttpVerb v : HttpVerb.values()){
+                    if (v.query.equals(verb)){
```
heyzooi: I'm wondering if this check needs to be case insensitive

android-lib/src/main/java/com/kinvey/android/async/AsyncBatchPushRequest.java
b-stolyarov: It will be a separated task for this feature.

android-lib/src/main/java/com/kinvey/android/async/AsyncBatchPushRequest.java
b-stolyarov: for `Kotlin` classes implementation,
Kotlin support updates currently only exists in `kotlin-migration` branch and that changes I think we cannot merge into `indev` branch because we need some tasks to do for full `Kotlin` support.
So currently I use `Java` for this classes. Maybe later we will convert all this classes to `Kotlin` but in this `kotlin-migration` branch then.

java-api-core/src/com/kinvey/java/model/KinveyBatchInsertError.java
b-stolyarov: fixed

java-api-core/src/com/kinvey/java/model/KinveyUpdateSingleItemError.java
b-stolyarov: Fixed.

java-api-core/src/com/kinvey/java/sync/dto/SyncRequest.java
b-stolyarov: Added fix to make this compare logic case insensitive.

android-lib/src/main/java/com/kinvey/android/async/AsyncBatchPushRequest.java
vinaygahlawat: This work is tracked by https://kinvey.atlassian.net/browse/MLIBZ-3089.

java-api-core/src/com/kinvey/java/model/KinveyUpdateSingleItemError.java
heyzooi: I see you fixed the other issues, but this one still as it was. maybe you forgot to commit this file

PR #239, opened 2019-07-09
Kotlin migration

b-stolyarov:
#### Description
Kotlin migration

#### Changes
Updated Gradle version to 5.1.1,
Support for Kotlin added
Changed JavaDoc generating tool from Doclava to Dokka 

#### Tests
Instrumented


android-lib/build.gradle
```diff
@@ -54,19 +54,27 @@ android {
     abortOnError false
   }
   packagingOptions {
+    exclude 'META-INF/NOTICE'
+    exclude 'META-INF/LICENSE'
```
heyzooi: @vinaygahlawat do we need to include the license files inside the package?

android-lib/build.gradle
```diff
@@ -54,19 +54,27 @@ android {
...
 
 configurations {
-  jaxDoclet
-  classpaths
   compile.exclude module: 'guava-jdk5'
 }
 
 buildscript {
+  ext.kotlin_version = '1.3.40'
+  ext.dokkaVersion = '0.9.17'
```
heyzooi: if the dokka also applies for the java-core project better to move the the `parent` gradle file

appcompat_v7/AndroidManifest.xml
```diff
@@ -1,20 +0,0 @@
...
-     Unless required by applicable law or agreed to in writing, software
-     distributed under the License is distributed on an "AS IS" BASIS,
-     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-     See the License for the specific language governing permissions and
-     limitations under the License.
--->
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
-          package="android.support.v7.appcompat">
-    <uses-sdk android:minSdkVersion="7"/>
-    <application />
```
heyzooi: do we need to have this `AndroidManifest.xml` since this is a library and not an application?

android-lib/build.gradle
b-stolyarov: Updated.

android-lib/build.gradle
vinaygahlawat: @heyzooi @b-stolyarov I think that we should include the license files inside the package.

b-stolyarov: Updated

PR #240, opened 2019-07-12
MLIBZ-2859: Client.getActiveUser() returns only userId

b-stolyarov:
#### Description
If login to an app, then to close the app and open it again `client.getActiveUser()` returns only user id, even if user contained some additional filled fields (userName etc). 

#### Changes
Added saving user info to realm table, except auth_token. 

#### Tests
Instrumented

### DIFF
--- a/java-api-core/src/com/kinvey/java/store/UserStoreRequestManager.java
+++ b/java-api-core/src/com/kinvey/java/store/UserStoreRequestManager.java
@@ -63,7 +63,7 @@ import java.util.Map;
 public class UserStoreRequestManager<T extends BaseUser> {
 
 
-    public static final String USER_COLLECTION_NAME = "user";
+    public static final String USER_COLLECTION_NAME = "active_user_info";
     public static final String GRANT_TYPE = "grant_type";
     public static final String USERNAME_PARAM = "username";
     public static final String PASSWORD_PARAM = "password";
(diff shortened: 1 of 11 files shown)


java-api-core/src/com/kinvey/java/store/UserStoreRequestManager.java
```diff
@@ -57,13 +58,15 @@
...
+import static com.kinvey.java.Constants.ACCESS_ERROR;
+
 /**
  * Created by Prots on 2/12/16.
  */
 public class UserStoreRequestManager<T extends BaseUser> {
 
 
-    public static final String USER_COLLECTION_NAME = "user";
+    public static final String USER_COLLECTION_NAME = "active_user_info";
```
heyzooi: any special reason why you need to change the collection name for the local cache?

heyzooi: I'm wondering if it would cause a breaking change

b-stolyarov: `USER_COLLECTION_NAME` wasn't used in the SDK, but I agree that it's public constant, I'll create `ACTIVE_USER_COLLECTION_NAME` for `active_user_info` and return back `user`.

b-stolyarov: I wouldn't like to use `user` for keeping user info, because this table name potentially can be busy.

b-stolyarov: Updated.

PR #243, opened 2019-07-25
MLIBZ-3100: Convert multi-insert code to Kotlin

b-stolyarov:
#### Description
Converting multi-insert code to `Kotlin`.

#### Changes
Converted multi-insert classes code to `Kotlin` language.

#### Tests
Instrumented.


android-lib/src/androidTest/java/com/kinvey/androidTest/store/data/DataStoreMultiInsertTest.java
```diff
@@ -1,1980 +0,0 @@
-package com.kinvey.androidTest.store.data;
```
heyzooi: I'm thinking if we should keep the unit tests in Java for now so we can assure the code is behaviour just like before the Kotlin migration

vinaygahlawat: @heyzooi +1. We should not migrate tests written in Java until we have completed the Kotlin migration and verify that all existing tests still pass.

b-stolyarov: Reverted changes of old `DataStoreMultiInsertTest.java` tests file, keep new `DataStoreMultiInsertTest.kt` tests file but set it as `Ignored` when tests run.

PR #245, opened 2019-07-31
MLIBZ-3089: Send multiple multi-insert requests when entity count exceeds max

b-stolyarov:
#### Description
Added logic to send multiple multi-insert requests when entity count exceeds max.

#### Changes
Added logic to send multiple multi-insert requests when entity count exceeds max.

#### Tests
Instrumented.


android-lib/src/androidTest/java/com/kinvey/androidTest/network/MockMultiInsertNetworkManager.java
```diff
@@ -36,14 +37,20 @@ public Save saveBlocking(T entity) throws IOException {
     @Override
     public SaveBatch saveBatchBlocking(List<T> list) throws IOException {
         useMultiInsertSave = true;
+        multiPostCount++;
```
heyzooi: I'm wondering if this method needs to be thread-safe

android-lib/src/androidTest/java/com/kinvey/androidTest/network/MockMultiInsertNetworkManager.java
```diff
@@ -36,14 +37,20 @@ public Save saveBlocking(T entity) throws IOException {
...
         return super.saveBatchBlocking(list);
     }
 
     public void clear() {
+        multiPostCount = 0;
```
heyzooi: also wondering if this method needs to be thread-safe

android-lib/src/androidTest/java/com/kinvey/androidTest/store/data/DataStoreMultiInsertTestKotlin.kt
```diff
@@ -0,0 +1,1954 @@
...
+            })
+            looperThread.start()
+        } else {
+            latch.countDown()
+        }
+        latch.await()
+        looperThread?.mHandler?.sendMessage(Message())
+    }
+
+    private class DefaultKinveyClientCallback<T : GenericJson> (private val latch: CountDownLatch) : KinveyClientCallback<T> {
```
heyzooi: is possible to create static classes in Kotlin?

android-lib/src/androidTest/java/com/kinvey/androidTest/store/data/DataStoreMultiInsertTestKotlin.kt
```diff
@@ -0,0 +1,1954 @@
...
+            this.error = error
+            finish()
+        }
+
+        internal fun finish() {
+            latch.countDown()
+        }
+    }
+
+    private class DefaultKinveyClientListCallback<T : GenericJson> (private val latch: CountDownLatch) : KinveyClientCallback<List<T>> {
```
heyzooi: these callbacks are better to be `protected` instead of `private`, so Mockito can mock those classes in runtime

android-lib/src/androidTest/java/com/kinvey/androidTest/store/data/DataStoreMultiInsertTestKotlin.kt
```diff
@@ -0,0 +1,1954 @@
...
+            latch.countDown()
+        }
+    }
+
+    private fun createPerson(name: String): Person {
+        return Person(name)
+    }
+
+    @Throws(InterruptedException::class)
+    private fun <T : GenericJson> delete(store: DataStore<T>, query: Query): DefaultKinveyDeleteCallback {
```
heyzooi: maybe better to move those methods like `delete`, `find`, `sync`, etc to a base class so it can be used by other tests in the future

android-lib/src/androidTest/java/com/kinvey/androidTest/store/data/DataStoreMultiInsertTestKotlin.kt
```diff
@@ -0,0 +1,1954 @@
...
+        val callback = DefaultKinveyClientCallback<T>(latch)
+        val looperThread = LooperThread(Runnable { store.save(item, callback) })
+        looperThread.start()
+        latch.await()
+        looperThread.mHandler.sendMessage(Message())
+        return callback
+    }
+
+    @Throws(InterruptedException::class)
+    private fun <T : GenericJson> push(store: DataStore<T>, seconds: Int): DefaultKinveyPushCallback {
```
heyzooi: let's move those methods to a base class and try to keep only methods annotated with `@Test` here

android-lib/src/main/java/com/kinvey/android/store/DataStore.java
```diff
@@ -578,8 +578,8 @@ private void saveBatch(@NonNull List<T> entities, @NonNull KinveyClientCallback<
         Preconditions.checkArgument(client.isInitialize(), "client must be initialized.");
         Preconditions.checkNotNull(entities, "Entity cannot be null.");
         Preconditions.checkState(entities.size() > 0,"Entity list cannot be empty.");
-        Preconditions.checkPositionIndex(entities.size(), MAX_MULTI_INSERT_SIZE,
-             String.format(Locale.US, "Reached maximum of %d items per request.", MAX_MULTI_INSERT_SIZE));
+        //Preconditions.checkPositionIndex(entities.size(), MAX_MULTI_INSERT_SIZE,
+        //     String.format(Locale.US, "Reached maximum of %d items per request.", MAX_MULTI_INSERT_SIZE));
```
heyzooi: I think better remove than comment in this case

java-api-core/src/com/kinvey/java/Constants.java
```diff
@@ -17,6 +17,9 @@
     public static final String AUTH_TOKEN = "authtoken";
     public static final String _KMD = "_acl";
     public static final String _ACL = "_acl";
+
+    public static final String META_ID = "meta.id";
```
heyzooi: I'm trying to find where this value is used and why we need it

java-api-core/src/com/kinvey/java/model/KinveyBatchInsertError.kt
```diff
@@ -0,0 +1,12 @@
+package com.kinvey.java.model
+
+import com.google.api.client.util.Key
+
+data class KinveyBatchInsertError (
```
heyzooi: I'm wondering if we need to extends from `GenericJson`, I guess not right?

java-api-core/src/com/kinvey/java/model/KinveySyncSaveBatchResponse.kt
```diff
@@ -0,0 +1,8 @@
+package com.kinvey.java.model
+
+import com.google.api.client.json.GenericJson
+import com.google.api.client.util.Key
+
+class KinveySyncSaveBatchResponse<T>(
+    var entityList: Collection<T?>?,
+    var errors: List<KinveyBatchInsertError>?) : GenericJson()
```
heyzooi: this one extends from `GenericJson` for example. I'm just trying to understand when we need and when we don't need to extends from `GenericJson`

android-lib/src/androidTest/java/com/kinvey/androidTest/network/MockMultiInsertNetworkManager.java
b-stolyarov: Created as atomic variable to make sure that is thread safe.

android-lib/src/androidTest/java/com/kinvey/androidTest/network/MockMultiInsertNetworkManager.java
b-stolyarov: Created as atomic variable to make sure that is thread-safe.

android-lib/src/androidTest/java/com/kinvey/androidTest/store/data/DataStoreMultiInsertTestKotlin.kt
b-stolyarov: If we need this to prevent memory-leak then in `Kotlin` if we just create an `inner` class by default that it's not keeps link to `outer` class and we can use it safely. So we can keep changes as it is.

android-lib/src/androidTest/java/com/kinvey/androidTest/store/data/DataStoreMultiInsertTestKotlin.kt
b-stolyarov: Changes callbacks modifiers to `protected`.

android-lib/src/androidTest/java/com/kinvey/androidTest/store/data/DataStoreMultiInsertTestKotlin.kt
b-stolyarov: Moved common code and these methods to base class `BaseDataStoreTest`.

android-lib/src/androidTest/java/com/kinvey/androidTest/store/data/DataStoreMultiInsertTestKotlin.kt
b-stolyarov: Moved common code base class `BaseDataStoreTest`.

android-lib/src/main/java/com/kinvey/android/store/DataStore.java
b-stolyarov: Fixed.

java-api-core/src/com/kinvey/java/Constants.java
b-stolyarov: We use this key in classes: `AsyncBatchPushRequest`, `PushBatchRequest`.

java-api-core/src/com/kinvey/java/model/KinveyBatchInsertError.kt
b-stolyarov: Yes, I used extending from `GenericJson` for response classes in my last implementation of methods for multi-insert in `SyncManager`. Now it's none needed. Removed.

java-api-core/src/com/kinvey/java/model/KinveySyncSaveBatchResponse.kt
b-stolyarov: Yes, I used extending from `GenericJson` for response classes in my last implementation of methods for multi-insert in `SyncManager`. Now it's none needed. Removed.

PR #246, opened 2019-07-31
MLIBZ-3091: Add Android Lint to project

b-stolyarov:
#### Description
Enabled Android Lint checks in project.

#### Changes
Enabled Android Lint checks in project.

#### Tests
Instrumented.

### DIFF
--- a/android-lib/src/main/java/com/kinvey/android/UuidFactory.java
+++ b/android-lib/src/main/java/com/kinvey/android/UuidFactory.java
@@ -24,8 +24,10 @@ import com.google.gson.Gson;
 
 import android.content.Context;
 import android.content.SharedPreferences;
+import android.content.pm.PackageManager;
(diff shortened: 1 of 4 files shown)


android-lib/src/main/java/com/kinvey/android/UuidFactory.java
```diff
@@ -88,28 +89,31 @@ public UuidFactory(Context context) {
...
                     }
-
                 }
             }
         }
-
     }
 
-
+    private String phoneDeviceId(Context context) {
```
vinaygahlawat: @b-stolyarov How are these changes for the device Id related to the Android Lint task?

b-stolyarov: This changes is to fix error in `Lint` report, that fails build process.

PR #247, opened 2019-07-31
MLIBZ-3071: Kotlin File Migration

heyzooi:
#### Description

Migrating all file related classes and interfaces

#### Changes

- FileStore and base classes
- File download and upload listeners
- Download and upload request classes

#### Tests

- Same unit tests


java-api-core/src/com/kinvey/java/network/NetworkFileManager.java
```diff
@@ -729,7 +736,7 @@ private DownloadMetadataAndFile(FileMetaData meta, DownloaderProgressListener pr
 
         @Key("id")
         private String id;
-        
+
         @Key("query")
         private String queryFilter;
         @Key("sort")
```
b-stolyarov: I think for `inner` classes in `Java` better use `inner static` classes, e.g.:
 ```public static class UploadMetadataAndFile...```
To avoid possibility of memory leaks.

Or maybe better convert them to `Kotlin` later where cannot be such problem if we declare classes that way.

heyzooi:
@b-stolyarov working on it... weird that running the unit tests locally those errors are not happening

heyzooi:
@b-stolyarov ignore my last comment, I'm now able to reproduce the error

java-api-core/src/com/kinvey/java/core/MetaUploadProgressListener.kt
```diff
@@ -14,43 +14,39 @@
...
+     * @param meta the metadata object of the file being downloaded.
      */
-    public FileMetaData getMetadata(){
-        return metadata;
+    fun metaDataRetrieved(meta: FileMetaData) {
+        //        this.metadata = meta;
+        this.metadata = FileMetaData()
+        this.metadata!!.fileName = meta.fileName
+        this.metadata!!.id = meta.id
+        Logger.INFO("cache meta: " + this.metadata!!.fileName + " and id: " + this.metadata!!.id)
```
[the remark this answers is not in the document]
heyzooi: fixed

java-api-core/src/com/kinvey/java/network/NetworkFileManager.java
heyzooi: just converted the entire class to Kotlin, since it's also part of the file group of classes

PR #248, opened 2019-08-02
MLIBZ-3110: AndroidX Migration

heyzooi:
#### Description

AndroidX is the new recommended package for a bunch of libraries that we use.

#### Changes

- Replace deprecated libraries in favor of `AndroidX` versions

#### Tests

- Same tests

### DIFF
--- a/android-lib/build.gradle
+++ b/android-lib/build.gradle
@@ -24,7 +24,7 @@ android {
 
   defaultConfig {
     minSdkVersion 16
-    targetSdkVersion 28
+    targetSdkVersion 29
     multiDexEnabled true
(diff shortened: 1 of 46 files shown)


android-lib/build.gradle
```diff
@@ -76,6 +76,7 @@ configurations {
 
 buildscript {
   ext.kotlin_version = '1.3.41'
+  ext.androidx_version = '1.3.0-alpha02'
```
vinaygahlawat: @heyzooi @b-stolyarov Do you have any concerns about using an `alpha` version of `androidx`?

heyzooi: right, better to change to the latest stable version, which is `1.2.0`

PR #254, opened 2019-08-09
MLIBZ-3073: Implement File upload

b-stolyarov:
#### Description
Implementation of file upload logic in `Kotlin`.

#### Changes
* Converted file upload classes to `Kotlin`.

#### Tests
Instrumented.


java-api-core/src/com/kinvey/java/network/NetworkManager.java
```diff
@@ -1015,7 +1015,7 @@ public T execute() throws IOException {
     }
 
     /** used internally **/
-    enum SaveMode {
+    public enum SaveMode {
```
vinaygahlawat: if this is used internally (as per the code comment), why is the access set to `public`?

b-stolyarov: Moved `SaveMode` enum class to separated file.

PR #258, opened 2019-08-20
MLIBZ-3123: Convert model classes

b-stolyarov:
#### Description
Converted model classes in the SDK to `Kotlin`.

#### Changes
* Converted all classes that in `model` folder of the SDK to `Kotlin`.

#### Tests
Instrumented.


android-lib/src/androidTest/java/com/kinvey/androidTest/store/datastore/CustomEndpointsTest.java
```diff
@@ -132,7 +132,7 @@ public void testCustomEndpointsNotExist() throws IOException, InterruptedExcepti
     */
     @Test
     public void testAsyncCustomEndpoints() throws IOException, InterruptedException {
-        AsyncCustomEndpoints endpoints = client.customEndpoints(GenericJson.class);
+        AsyncCustomEndpoints endpoints = (AsyncCustomEndpoints) client.customEndpoints(GenericJson.class);
```
heyzooi: why this cast is now necessary?

android-lib/src/main/java/com/kinvey/android/Client.kt
```diff
@@ -0,0 +1,1136 @@
...
+     * Sample Usage:
+     * <pre>
+     * `CustomEndpoints myCustomEndpoints = kinveyClient.customEndpoints();
+    ` *
+    </pre> *
+     *
+     *
+     * @return Instance of [com.kinvey.java.UserDiscovery] for the defined collection
+     */
+    @Deprecated("")
```
heyzooi: let's add some useful message why this is now deprecated

android-lib/src/main/java/com/kinvey/android/async/AsyncBatchPushRequest.kt
```diff
@@ -130,7 +130,7 @@ class AsyncBatchPushRequest<T : GenericJson>(
                         manager.deleteCachedItem(syncItem[_ID] as String?)
                     }
                 } catch (e: IOException) {
-                    val err = KinveyUpdateSingleItemError(e, item)
+                    val err = KinveyUpdateSingleItemError(e, item as GenericJson)
```
heyzooi: wondering why we need a casting here

android-lib/src/main/java/com/kinvey/android/callback/KinveyUserCallback.kt
```diff
@@ -28,7 +28,7 @@ import com.kinvey.java.store.UserStoreRequestManager
 interface KinveyUserCallback<T : User> : KinveyClientCallback<T> {
 
 
-    override fun onSuccess(result: T)
+    override fun onSuccess(result: T?)
```
heyzooi: success responses should always have a value, if the response is `void` I believe the `Nothing` Kotlin keyword can be used

java-api-core/src/com/kinvey/java/model/AggregateType.kt
```diff
@@ -1,10 +1,10 @@
-package com.kinvey.java.model;
+package com.kinvey.java.model
 
 /**
  * Created by yuliya on 10/12/17.
  */
 
-public enum AggregateType {
```
heyzooi: I'm thinking if we should turn this private, if so it would be a breaking change

java-api-core/src/com/kinvey/java/model/KinveySyncSaveBatchResponse.kt
```diff
@@ -1,5 +1,6 @@
 package com.kinvey.java.model
 
-class KinveySyncSaveBatchResponse<T>(
+data class KinveySyncSaveBatchResponse<T>(
```
heyzooi: sorry about my lack of better knowledge about Kotlin, but what's the difference between a regular `class` and `data class`?

java-api-core/src/com/kinvey/java/store/requests/data/save/SaveListBatchRequest.kt
```diff
@@ -87,7 +87,7 @@ class SaveListBatchRequest<T : GenericJson>(
         val batchSaveErrors = ArrayList<KinveyBatchInsertError>()
 
         val updateResponse = updateObjects(updateList)
-        saveEntities.addAll(updateResponse.entities)
+        saveEntities.addAll(updateResponse.entities as Collection<T>)
```
heyzooi: is this cast required?

android-lib/src/androidTest/java/com/kinvey/androidTest/store/datastore/CustomEndpointsTest.java
b-stolyarov: Fixed.

android-lib/src/main/java/com/kinvey/android/Client.kt
b-stolyarov: Added text to point user to use new method.

android-lib/src/main/java/com/kinvey/android/async/AsyncBatchPushRequest.kt
b-stolyarov: Removed cast.

android-lib/src/main/java/com/kinvey/android/callback/KinveyUserCallback.kt
b-stolyarov: Removed nullable `T?` type here.
Added for code where entity can be `null` some additional checks.

java-api-core/src/com/kinvey/java/model/KinveySyncSaveBatchResponse.kt
b-stolyarov: Data classes are designed to store data, can be useful for network request entities.
This type of classes can automatically create equals and hashCode implementation with variables that in class constructor.
If we use class for store or transfer data then preferable to use data classes instead of regular classes.
So I used `data class` in such cases.

java-api-core/src/com/kinvey/java/store/requests/data/save/SaveListBatchRequest.kt
b-stolyarov: Added additional conversion for entities list to make it immutable.

java-api-core/src/com/kinvey/java/model/AggregateType.kt
b-stolyarov: By default, if no access modifier placed for `Kotlin` class it uses public access.

PR #259, opened 2019-08-22
MLIBZ-3122: Convert all query classes.

b-stolyarov:
#### Description
Converted query classes to `Kotlin`.

#### Changes
* Converted request classes and query related classes to `Kotlin`.

#### Tests
Instrumented.


android-lib/src/androidTest/java/com/kinvey/androidTest/store/datastore/CustomEndpointsTest.java
```diff
@@ -132,7 +132,7 @@ public void testCustomEndpointsNotExist() throws IOException, InterruptedExcepti
     */
     @Test
     public void testAsyncCustomEndpoints() throws IOException, InterruptedException {
-        AsyncCustomEndpoints endpoints = client.customEndpoints(GenericJson.class);
+        AsyncCustomEndpoints endpoints = (AsyncCustomEndpoints) client.customEndpoints(GenericJson.class);
```
heyzooi: samea as the other PR, I don't think this cast is necessary

android-lib/src/main/java/com/kinvey/android/async/AsyncBatchPushRequest.kt
```diff
@@ -130,7 +130,7 @@ class AsyncBatchPushRequest<T : GenericJson>(
                         manager.deleteCachedItem(syncItem[_ID] as String?)
                     }
                 } catch (e: IOException) {
-                    val err = KinveyUpdateSingleItemError(e, item)
+                    val err = KinveyUpdateSingleItemError(e, item as GenericJson)
```
heyzooi: since this PR seems to be part of the other one, maybe better to merge #258 first so we don't have to repeat the same comments here

android-lib/src/androidTest/java/com/kinvey/androidTest/store/datastore/CustomEndpointsTest.java
b-stolyarov: Merged approved changes from `master`.

java-api-core/src/com/kinvey/java/network/NetworkFileManager.kt
```diff
@@ -720,8 +720,8 @@ open class NetworkFileManager
 
         init {
             this.queryFilter = query.getQueryFilterJson(networkFileManager.client.jsonFactory)
-            val queryLimit = query.limit
-            val querySkip = query.skip
+            val queryLimit = query.getLimit()
+            val querySkip = query.getSkip()
```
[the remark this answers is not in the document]
b-stolyarov: Yes, right we can convert this to use property. Fixed.

PR #261, opened 2019-08-27
refactor: MLIBZ-3076, MIC Kotlin migration

heyzooi:
#### Description

MIC Kotlin migration

#### Changes

- `MIC` related classes

#### Tests

- Same tests


java-api-core/src/com/kinvey/java/core/AbstractKinveyClientRequest.kt
```diff
@@ -49,7 +49,7 @@ protected constructor(
         /**
          * Kinvey JSON client *
          */
-        private val client: AbstractClient<BaseUser>, private val hostName: String,
+        private val client: AbstractClient<*>, private val hostName: String,
```
vinaygahlawat: What are the implications of changing the template type here? Do we have another type that `AbstractClient` can contain?

PR #272, opened 2019-09-20
MLIBZ-3140: datastore.save failing for network error invokes the onFailure handler - Android

b-stolyarov:
#### Description
Fix for `DataStore.save` failing for network error invokes the `onFailure` handler.
#### Changes
* Fix for `DataStore.save` failing for network error invokes the `onFailure` handler, now it returns onSuccess if item was saved in cache if was no connection, for `AUTO` or `SYNC` store types.
* Made some correction in tests: `DataStoreTest#testSaveConnectErrorAuto`, `DataStoreTest#testSaveNoAccess` that checking old `DataStore.save` behavior, that was changed.
* Added new tests to check new `DataStore.save` logic behaviour.
#### Tests
Instrumented.

### DIFF
--- a/java-api-core/src/com/kinvey/java/store/requests/data/save/SaveRequest.kt
+++ b/java-api-core/src/com/kinvey/java/store/requests/data/save/SaveRequest.kt
@@ -71,12 +71,12 @@ class SaveRequest<T : GenericJson>(private val cache: ICache<T>?, private val ne
                         // to remove the entity from the local cache with the temporary ID.
                         cache?.delete(id)
                     }
+                    ret?.let { cache?.save(it) }
                 } catch (e: IOException) {
                     syncManager?.enqueueRequest(networkManager?.collectionName,
                     networkManager, if (bRealmGeneratedId == true) SyncRequest.HttpVerb.POST else SyncRequest.HttpVerb.PUT, id)
-                    throw e
+                    //throw e
                 }
-                ret?.let { cache?.save(it) }
             }
             WritePolicy.FORCE_NETWORK -> {
                 Logger.INFO("Start saving entity")
--- a/java-api-core/src/com/kinvey/java/store/requests/data/save/SaveRequest.kt
+++ b/java-api-core/src/com/kinvey/java/store/requests/data/save/SaveRequest.kt
@@ -16,6 +16,7 @@
 
 package com.kinvey.java.store.requests.data.save
 
+import com.google.api.client.http.HttpResponseException
 import com.google.api.client.json.GenericJson
 import com.kinvey.java.Constants
 import com.kinvey.java.Logger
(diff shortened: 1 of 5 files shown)


java-api-core/src/com/kinvey/java/store/requests/data/save/SaveRequest.kt
```diff
@@ -71,12 +71,12 @@ class SaveRequest<T : GenericJson>(private val cache: ICache<T>?, private val ne
                         // to remove the entity from the local cache with the temporary ID.
                         cache?.delete(id)
                     }
+                    ret?.let { cache?.save(it) }
                 } catch (e: IOException) {
                     syncManager?.enqueueRequest(networkManager?.collectionName,
                     networkManager, if (bRealmGeneratedId == true) SyncRequest.HttpVerb.POST else SyncRequest.HttpVerb.PUT, id)
-                    throw e
+                    //throw e
```
heyzooi: check with @Martin-Hristov but I believe you also need to check if the http response code is `401`, if so, the error should be throw

java-api-core/src/com/kinvey/java/store/requests/data/save/SaveRequest.kt
```diff
@@ -26,8 +27,10 @@ import com.kinvey.java.store.requests.data.IRequest
 import com.kinvey.java.store.requests.data.PushRequest
 import com.kinvey.java.sync.SyncManager
 import com.kinvey.java.sync.dto.SyncRequest
+import org.apache.http.HttpException
 
 import java.io.IOException
+import java.net.UnknownHostException
```
heyzooi: those imports does not seems necessary

java-api-core/src/com/kinvey/java/store/requests/data/save/SaveRequest.kt
```diff
@@ -71,12 +74,12 @@ class SaveRequest<T : GenericJson>(private val cache: ICache<T>?, private val ne
                         // to remove the entity from the local cache with the temporary ID.
                         cache?.delete(id)
                     }
+                    ret?.let { cache?.save(it) }
                 } catch (e: IOException) {
                     syncManager?.enqueueRequest(networkManager?.collectionName,
                     networkManager, if (bRealmGeneratedId == true) SyncRequest.HttpVerb.POST else SyncRequest.HttpVerb.PUT, id)
-                    throw e
+                    if (e is HttpResponseException && e.statusCode == 401) throw e
```
heyzooi: let's wrap the original exception inside another exception object so we don't lose the original stack details.
something like `throw new IOException(e)`, I think that's the Java way of doing this since Java does not have a better way to rethrow an exception like C# does

PR #279, opened 2020-03-06
MLIBZ-3048 Multi record insert create

b-stolyarov:
#### Description
Adding new create multiinsert method

#### Changes
Added create method only for kinvey api version 5

#### Tests
Instrumented


dimofeev:
@b-stolyarov can we proceed here?

PR #280, opened 2020-06-02
KDEV 547: Return items count with GET requests

b-stolyarov:
#### Description
Add an option to send 'X-Kinvey-Include-Items-Count' header and get the returned items in 'X-Kinvey-Items-Count' returned header.

#### Changes
- Added `isAddCountHeader` parameter to `DataStore`. `isAddCountHeader` is false by default. If `isAddCountHeader` is true, all callbacks `KinveyReadCallback` have not null `count` field with count of all items by a query. `limit` and `skip` query parameters are ignored.
- Added `findWithCount(query: Query, callback: KinveyReadCallback<T>)` method to `DataStore`. The method returns count of all items by the query in  `count` field in `KinveyReadCallback`. This method ignores `isAddCountHeader` parameter from `DataStore`.
- Added `count` field to `KinveyReadCallback` which shows count of all items in collection for a query.  If `isAddCountHeader` is false or `findWithCount` isn't used, the `count` field is null.

#### Tests
Instrumented

### DIFF
--- a/android-lib/build.gradle
+++ b/android-lib/build.gradle
@@ -168,6 +168,8 @@ dependencies {
   // Multidex Support for testing
   androidTestImplementation group: 'androidx.multidex', name: 'multidex-instrumentation', version: '2.0.0'
   implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
+
+  androidTestImplementation 'androidx.test.ext:junit:1.1.1'
 }
 
 task superJar(type: Jar) {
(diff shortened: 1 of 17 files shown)


b-stolyarov:
> Shouldn't it be "hasAddCountHeader" and not "isAddCountHeader"? From what I remember back in my Java days, both "is" and "has" are fine for bool properties and you should select what makes sense. I think for this case it should be "has".

It Makes sense, let's only then change it to `hasCountHeader`. WDYT?

PR #285, opened 2020-06-26
KDEV-784: Create function should support also a single item

b-stolyarov:
#### Description
The `create` function should support also a single item.

#### Changes
Added `create(entity: T, callback: KinveyClientCallback<T>)` function to `DataStore` class.

#### Tests
Instrumented.

### DIFF
--- a/android-lib/src/androidTest/java/com/kinvey/androidTest/store/data/multiinsert/DataStoreSingleInsertTest.kt
+++ b/android-lib/src/androidTest/java/com/kinvey/androidTest/store/data/multiinsert/DataStoreSingleInsertTest.kt
@@ -3,20 +3,16 @@ package com.kinvey.androidTest.store.data.multiinsert
 import androidx.test.filters.LargeTest
 import androidx.test.runner.AndroidJUnit4
 import com.google.api.client.json.GenericJson
-import com.kinvey.android.Client
 import com.kinvey.android.store.DataStore
 import com.kinvey.androidTest.model.EntitySet
 import com.kinvey.androidTest.model.Person
-import com.kinvey.androidTest.network.MockMultiInsertNetworkManager
-import com.kinvey.java.KinveySaveBatchException
 import com.kinvey.java.core.KinveyJsonResponseException
 import com.kinvey.java.store.StoreType
-import com.kinvey.java.sync.dto.SyncRequest
 import com.kinvey.java.AbstractClient.Companion.kinveyApiVersion
 import com.kinvey.java.Constants._ID
+import org.junit.Assert
 
 
-import org.junit.Ignore
 import org.junit.Test
 import org.junit.runner.RunWith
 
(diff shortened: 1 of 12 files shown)


android-lib/src/androidTest/java/com/kinvey/androidTest/store/data/multiinsert/DataStoreSingleInsertTest.kt
```diff
@@ -0,0 +1,223 @@
...
+        clearBackend(netStore)
+        var saveCallback = createList(netStore, personList)
+        assertNull(saveCallback.error)
+        assertNotNull(saveCallback.result)
+        Assert.assertTrue(checkPersonIfSameObjects(personList, saveCallback.result?.entities, false))
+        saveCallback = createList(netStore, personList)
+        assertNull(saveCallback.error)
+        assertNotNull(saveCallback.result?.errors)
+        assertNotNull(saveCallback.result?.errors?.get(0)?.description)
+        assertNotNull(saveCallback.result?.errors?.get(0)?.debug)
```
alexanderfilipov, line 221: In general, I'd suggest to verify not just the field existence, but also the exact string, or at least some parts of it. For example, in this case the `description/debug` strings would contain something like "already exists" or "duplicate id', which we can compare against. 

I guess this is not done at the moment because theoreticaly the messages can be changed in the backend, but this is not something we usually do. And if we change something like this, it should be in a new API version. So the old one should continue to work as before, unless deprecated entirely.

android-lib/src/main/java/com/kinvey/android/store/DataStore.kt
```diff
@@ -418,6 +416,39 @@ open class DataStore<T : GenericJson> : BaseDataStore<T> {
...
+     *
+     *
+     * @param entity The entity to create
+     * @param callback KinveyClientCallback<T>
+    </T> */
+    fun create(entity: T, callback: KinveyClientCallback<T>) {
+        Preconditions.checkNotNull(client, "client must not be null")
+        Preconditions.checkArgument(client?.isInitialize ?: false, "client must be initialized.")
+        Preconditions.checkNotNull(entity, "Entity cannot be null.")
+        Preconditions.checkState(kinveyApiVersion >= KINVEY_API_VERSION_5, "Kinvey api version cannot be less than 5.")
```
alexanderfilipov, line 447: Actually for single-entity insert all api versions would work. This check should only be done when calling `create` with an array of entities.

android-lib/src/androidTest/java/com/kinvey/androidTest/store/data/multiinsert/DataStoreSingleInsertTest.kt
b-stolyarov: Fixed

android-lib/src/main/java/com/kinvey/android/store/DataStore.kt
b-stolyarov: Fixed

PR #290, opened 2020-08-05
fix: KDEV-957: Not able to create items more than once

b-stolyarov:
#### Description
The create operation fails with error `An entity with _id ...` if create several data items, using the new 'create' function and an 'Auto' store and then try to create another several data items, using the new 'create' function and an 'Auto' store.

#### Changes
Added additional checking before searching items with ID in local storage.

#### Tests
Instrumented.


alexanderfilipov (approved):
LGTM.

